PR #1, opened 2020-01-24
#37820 Invalid link causes error (actually a bug in AnchorTagHelper)

jamesgriff:


GenderPayGap.WebUI/Helpers/UriSanitiser.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace GenderPayGap.WebUI.Helpers
+{
+    public static class UriSanitiser
+    {
+
+        public static bool IsValidHttpOrHttpsLink(string uri)
+        {
+            bool uriIsValid = Uri.TryCreate(uri, UriKind.Absolute, out Uri uriResult)
+                              && (uriResult.Scheme == Uri.UriSchemeHttp || uriResult.Scheme == Uri.UriSchemeHttps);
```
raamSoftwire: That's a neat use of the `out` parameter

PR #2, opened 2020-01-24
#37486 remind users to submit

raamSoftwire:
Creates new webjob to send reminders to users who are registered for in-scope organisations that have not yet reported their GPG.

- Creates a new DB table `ReminderEmails`
- Creates a new webjob that runs every hour
- Adds a config setting to set when reminders should be sent


GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
```diff
@@ -0,0 +1,190 @@
...
+using GenderPayGap.Database.Models;
+using GenderPayGap.Extensions.AspNetCore;
+using Microsoft.Azure.WebJobs;
+using Newtonsoft.Json;
+
+namespace GenderPayGap.WebJob
+{
+    public partial class Functions
+    {
+        public void SendReminderEmails([TimerTrigger("0 * * * *")] TimerInfo timer)
```
jamesgriff: It might be nice to have a comment to say what this `0 * * * *` means.
I knew you wanted to have it run every hour, but the crontab syntax means I had to scratch my head for a while to check it was right

GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        private static DateTime GetLatestReminderEmailDate(SectorTypes sectorType)
+        {
+            List<int> reminderEmailDays = GetReminderEmailDays();
+            var deadlineDate = GetDeadlineDate(sectorType);
+            
+            IEnumerable<DateTime> reminderDates = reminderEmailDays.Select(r => deadlineDate.AddDays(-r));
+            var latestDate = reminderDates.FirstOrDefault(r => r < DateTime.Now);
+            return latestDate;
+        }
```
jamesgriff: It might be nice to create a method `GetReminderDates(SectorType)`, which does...
```
private static List<DateTime> GetReminderDates(SectorTypes sectorType)
{
    List<int> reminderDays = GetReminderEmailDays();
    var deadlineDate = GetDeadlineDate(sectorType);

    return reminderDays.Select(reminderDay => deadlineDate.AddDays(-reminderDay)).ToList();
}
```
...(mostly copied from `GetLatestReminderEmailDate`)

Then we can just do
`GetReminderDates(sectorType).Min()`
or
```
GetReminderDates(sectorType)
    .Where(reminderDate => reminderDate < DateTime.Now)
    .OrderBy(reminderDate => reminderDate)
    .FirstOrDefault()
```

GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
raamSoftwire: I agree, I've added one in now

GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
raamSoftwire: That sounds good, thanks

raamSoftwire:
I've added a few logs, and I think we can get the rest of the information out of the database. I think the last thing to do is to update the emails on the DEV DB to use the smoke-test address (and set a few to my own address to see some come through)

```
UPDATE dbo.Users 
SET EmailAddress = 'simulate-delivered@notifications.service.gov.uk'
```

jamesgriff:
I added a comment to think about here:
https://github.com/cabinetoffice/gender-pay-gap/commit/012a89e27ebbfce800d6cdf8e8b09ab1bcb67b73

But otherwise, I'm happy to test this on Dev, as you describe

GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
```diff
@@ -21,50 +21,51 @@ public void SendReminderEmails([TimerTrigger("0 * * * *")] TimerInfo timer)
         {
             var start = DateTime.Now;
             CustomLogger.Information("SendReminderEmails Function started", start);
+            
+            try
             {
```
raamSoftwire: There was an extra layer of nesting that I think was unnecessary, so I've removed it. I

jamesgriff: Looks good :-)

PR #5, opened 2020-01-24
#38878 On logout, manually expire cookies

jamesgriff:
Make sure sign-out still works with Chrome's new cookie settings


raamSoftwire:
Looks good. Thanks James

PR #6, opened 2020-01-27
#39113 Replace Datetime with VirtualDateTime

raamSoftwire:


WarmUp/Program.cs
```diff
@@ -5,6 +5,7 @@
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using GenderPayGap.Extensions;
```
jamesgriff: The WarmUp script is its own, entirely separate application.
We run it on the build server, not in a context that has the configuration necessary to change the VirtualDateTime.
So, for WarmUp, I think we shouldn't use VirtualDateTime.

WarmUp/WarmUp.csproj
```diff
@@ -5,4 +5,8 @@
     <TargetFramework>netcoreapp2.2</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\GenderPayGap.Extensions\GenderPayGap.Extensions.csproj" />
+  </ItemGroup>
+
```
jamesgriff: Similarly to my other comment, I think WarmUp shouldn't depend on the other projects.

WarmUp/Program.cs
raamSoftwire: Good point, I've reverted these changes and force pushed

WarmUp/WarmUp.csproj
raamSoftwire: Fixed

jamesgriff:
@raamSoftwire Once you're done testing the Reminder Email ticket, remember to change that class to use `VirtualDateTime`

raamSoftwire:
Will do, I've made a note to sort this

PR #9, opened 2020-01-28
Fix bug in admin change scope page (also allow admins to change the scope for any year, not just current year)

jamesgriff:


GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationScopeController.cs
```diff
@@ -103,18 +110,19 @@ public async Task<IActionResult> ChangeScopePost(long id, AdminChangeScopeViewMo
...
         }
 
-        private void RetireOldScopesForCurrentSnapshotDate(Organisation organisation)
+        private void RetireOldScopesForCurrentSnapshotDate(Organisation organisation, int year)
         {
-            IOrderedEnumerable<OrganisationScope> organisationScopesForCurrentSnapshotDate = organisation.OrganisationScopes
-                .OrderByDescending(s => s.SnapshotDate);   
+            IEnumerable<OrganisationScope> organisationScopesForCurrentSnapshotDate =
+                organisation.OrganisationScopes
+                .Where(scope => scope.SnapshotDate.Year == year);
```
jamesgriff: This was the bug - we were setting all scope (for all years) to Retired.
Now we're only setting all scopes for the current year to Retired

raamSoftwire: Good spot. Thanks for fixing this

GenderPayGap.WebUI/Views/Admin/ViewOrganisation.cshtml
```diff
@@ -179,10 +179,19 @@
...
-                @(Model.LatestScope.ScopeStatus)
+                @{
+                    var scopes = Model.OrganisationScopes
+                        .Where(s => s.Status == ScopeRowStatuses.Active)
+                        .OrderByDescending(s => s.SnapshotDate);
+                }
+                @foreach (var scope in scopes)
+                {
+                    <b>
+                        @($"{scope.SnapshotDate.Year}-{(scope.SnapshotDate.Year + 1) % 100}"):
```
raamSoftwire: This is neat, but took me a second to figure out what is going on. Could we add a comment, or pull out a method for this formatting with a nice, descriptive method name?

raamSoftwire: I looks like we would re-use it a few times, so maybe a method over a comment makes sense?

PR #11, opened 2020-01-29
#37486 Fix SendReminderEmail Function

raamSoftwire:
The function would start, successfully send the first email to me (valid email), then fall over on the second one which was an invalid email (Notify smoke test address). This would result in only a single result in the DB.

The problem was with our CustomLogger. I was passing a `User` object, which it then tried to convert to JSON, but it would go and get the associated `UserOrganisations` which would in turn go and get the `User`. This caused the program to hang. This also explains why in the Azure logs we would see the function start but never finish.

I've also taken this opportunity to remove some of the other logging and try/catch blocks. Hopefully making the code more readable.

### DIFF
--- a/GenderPayGap.Core/Classes/Logger/CustomLogger.cs
+++ b/GenderPayGap.Core/Classes/Logger/CustomLogger.cs
@@ -35,6 +35,8 @@ namespace GenderPayGap.Core.Classes.Logger
         {
             try
             {
+                // The logger doesn't use JsonConvert but it is doing something similar
+                // We try to convert values to JSON to see if it will throw an exception
                 JsonConvert.SerializeObject(values);
                 return values == null ? message : message + ". Log: {@Values}";
             }
--- a/GenderPayGap.Core/Classes/Logger/CustomLogger.cs
+++ b/GenderPayGap.Core/Classes/Logger/CustomLogger.cs
@@ -1,4 +1,5 @@
-﻿using Serilog;
+﻿using Newtonsoft.Json;
+using Serilog;
 
 namespace GenderPayGap.Core.Classes.Logger
 {
@@ -32,7 +33,16 @@ namespace GenderPayGap.Core.Classes.Logger
 
         private static string GetLogMessage(string message, object values = null)
         {
-            return values == null ? message : message + ". Log: {@Values}";
+            try
+            {
+                JsonConvert.SerializeObject(values);
+                return values == null ? message : message + ". Log: {@Values}";
+            }
+            catch (JsonSerializationException ex)
+            {
+                Error("SERILOG ERROR: Can't serialize values");
+                throw;
+            }
         }
 
     }
--- a/GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
+++ b/GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
@@ -85,7 +85,7 @@ namespace GenderPayGap.WebJob
                     catch (Exception ex)
                     {
                         CustomLogger.Error(
-                            "Failed whilst sending or saving reminder email.",
+                            "Failed whilst sending or saving reminder email",
                             new
                             {
                                 UserId = user.UserId,
--- a/GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
+++ b/GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Autofac;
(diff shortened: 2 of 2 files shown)


GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
```diff
@@ -65,7 +65,6 @@ public void SendReminderEmails([TimerTrigger("0 * * * *")] TimerInfo timer)
             catch (Exception ex)
             {
                 CustomLogger.Error("Failed to send reminder emails", ex);
-                throw;
```
raamSoftwire: I think this was throwing the exception that crashed the webjob. The intended behaviour is that if one fails, it should log the problem but then move on to the next user. But I don't see this error message in the Azure logs...

raamSoftwire: But we do see the error on line 154,  "Failed to send reminder email to Notify.". That exception does get thrown, so I'd expect to end up in the catch block on line 95. But we don't see the "Failed whilst sending or saving reminder email." error message

jamesgriff: If you want to catch an error for each user individually, the try-catch will need to be within the for-each loop

GenderPayGap.Core/Classes/Logger/CustomLogger.cs
```diff
@@ -32,7 +33,16 @@ public static void Fatal(string message, object values = null)
 
         private static string GetLogMessage(string message, object values = null)
         {
-            return values == null ? message : message + ". Log: {@Values}";
+            try
+            {
+                JsonConvert.SerializeObject(values);
```
raamSoftwire: The logger doesn't use `JsonConvert` but our assumption is that it is doing something similar. This try/catch will stop us from getting caught out in future.

GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
```diff
@@ -23,51 +23,44 @@ public void SendReminderEmails([TimerTrigger("0 * * * *")] TimerInfo timer)
         {
             DateTime start = VirtualDateTime.Now;
             CustomLogger.Information("SendReminderEmails Function started", start);
+            
+            List<int> reminderDays = GetReminderEmailDays();
+            if (reminderDays.Count == 0)
+            {
+                CustomLogger.Information("SendReminderEmails Function finished. No ReminderEmailDays set.");
```
raamSoftwire: This is also a new log, because I got caught out once.

GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
```diff
@@ -87,44 +80,25 @@ public void SendReminderEmails([TimerTrigger("0 * * * *")] TimerInfo timer)
...
                         CustomLogger.Error(
                             "Failed whilst sending or saving reminder email.",
                             new
                             {
-                                User = user,
+                                UserId = user.UserId,
                                 SectorType = sectorType,
-                                Organisations = inScopeOrganisationsThatStillNeedToReport,
+                                OrganisationIds = inScopeOrganisationsThatStillNeedToReport.Select(o => o.OrganisationId),
                                 Exception = ex.Message
```
raamSoftwire: These were the lines of code that were causing the function to hang. The CustomLogger couldn't parse `user` or `inScopeOrganisationsThatStillNeedToReport` into JSON

GenderPayGap.Core/Classes/Logger/CustomLogger.cs
jamesgriff: Can you add a comment explaining this.
I'm worried a developer in the future will think they can remove this code because we're not using the value returned from `JsonConvert.SerializeObject(values)`

GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
jamesgriff: Does the `OrganisationIds` property look sensible in the logs?
I'm wondering because it's an `IEnumerable` and wondering if we need to do a `.ToList()` on it.

raamSoftwire: Great thinking and worth checking.

It looks fine 

`Failed whilst sending or saving reminder email.. Log: {"UserId": 13, "SectorType": "Public", "OrganisationIds": [12, 14], "Exception": "Status code 400.....`

GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
```diff
@@ -85,7 +85,7 @@ public void SendReminderEmails([TimerTrigger("0 * * * *")] TimerInfo timer)
                     catch (Exception ex)
                     {
                         CustomLogger.Error(
-                            "Failed whilst sending or saving reminder email.",
```
raamSoftwire: CustomLogger adds a full stop for us, so we were getting two.

PR #18, opened 2020-02-03
AB#39114 db integrity checks

raamSoftwire:
- Add new page to admin portal
- Runs database checks and displays results
- Uses AJAX calls to load results


GenderPayGap.WebUI/Views/AdminDatabaseIntegrityChecks/PublicSectorOrganisationsWithoutAPublicSectorType.cshtml
```diff
@@ -0,0 +1,45 @@
+﻿@using GenderPayGap.Database
+@model GenderPayGap.WebUI.Models.Admin.DatabaseIntegrityChecksViewModel
```
jamesgriff: The model here could probably be `List<Organisation>` - then you wouldn't have to do `.PublicSectorOrganisationsWithoutAPublicSectorType` loads of times

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
```diff
@@ -0,0 +1,201 @@
...
+            this.dataRepository = dataRepository;
+        }
+
+        [HttpGet("database-integrity-checks")]
+        public async Task<IActionResult> DatabaseIntegrityChecks()
+        {
+            return View("DatabaseIntegrityChecks");
+        }
+
+        [HttpGet("organisations-with-multiple-active-addresses-ajax")]
```
jamesgriff: It might be nice if the URL was `database-integrity-checks/organisations-with-multiple-active-addresses-ajax`, so it's clear that it's meant to be part of the DB integrity checks

GenderPayGap.WebUI/Models/Admin/DatabaseIntegrityChecksViewModel.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Collections.Generic;
```
jamesgriff: This file might not be needed - it might work if the model for each ajax call is just `List<Organisation>`

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
```diff
@@ -0,0 +1,201 @@
...
+        public async Task<IActionResult> OrganisationsWithMultipleActiveAddressesAjax()
+        {
+            var model = new DatabaseIntegrityChecksViewModel();
+            List<Organisation> organisations = dataRepository.GetAll<Organisation>()
+                .Include(o => o.OrganisationAddresses)
+                .ToList();
```
jamesgriff: It might be nice not to do `.ToList()` in all these methods.
That way, the results will be streamed from the database, rather than being read into memory all in one go.

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            foreach (Organisation organisation in organisations)
+            {
+                IEnumerable<OrganisationAddress> activeAddresses = organisation.OrganisationAddresses
+                    .Where(oa => oa.Status == AddressStatuses.Active);
+                if (activeAddresses.Count() > 2)
```
jamesgriff: (not important)
If you want, you can simplify
`enumerable.Where(x => x.Foo).Count()`
to
`enumerable.Count(x => x.Foo)`

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
```diff
@@ -0,0 +1,201 @@
...
+                    .Where(os => os.Status == ScopeRowStatuses.Active)
+                    .GroupBy(scope => scope.SnapshotDate)
+                    .Select(g => g.Key.Year)
+                    .ToList();
+
+                foreach (int year in requiredYears)
+                {
+                    if (!yearsWithActiveScopes.Contains(year))
+                    {
+                        model.OrganisationsWithNoActiveScopeForEveryYear.Add(organisation);
```
jamesgriff: This might add an organisation multiple times, if several years are missing

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
```diff
@@ -0,0 +1,201 @@
...
+                .ToList();
+
+            foreach (Organisation organisation in organisations)
+            {
+                List<Return> activeReturns = organisation.Returns
+                    .Where(r => r.Status == ReturnStatuses.Submitted)
+                    .ToList();
+
+                bool multipleActiveReturnsForYear = activeReturns
+                    .GroupBy(r => r.AccountingDate)
```
jamesgriff: Maybe even `.GroupBy(r => r.AccountingDate.Year)` to catch the case where there's an organisation with:
* 1 return for the public-sector date
* 1 return for the private-sector date

GenderPayGap.WebUI/Views/AdminDatabaseIntegrityChecks/DatabaseIntegrityChecks.cshtml
```diff
@@ -0,0 +1,195 @@
...
+        };
+    }
+
+    @(Html.GovUkBreadcrumbs(new BreadcrumbsViewModel
+    {
+        Crumbs = crumbs
+    }))
+}
+
+<main id="content" role="main">
```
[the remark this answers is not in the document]
raamSoftwire: Ahh so it is, I must have copied the format from a file that does it the old way. Thanks for pointing it out

GenderPayGap.WebUI/Views/AdminDatabaseIntegrityChecks/DatabaseIntegrityChecks.cshtml
```diff
@@ -0,0 +1,195 @@
...
+    <span class="govuk-caption-xl">Administration</span>
+    <h1 class="govuk-heading-xl">Database integrity checks</h1>
+
+    <div>
```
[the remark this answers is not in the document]
raamSoftwire: It was intended to group similar results together with the related title. It also makes the file more readable because of the added indentation. Would a more semantic `<section>` tag be better?

GenderPayGap.WebUI/Views/AdminDatabaseIntegrityChecks/DatabaseIntegrityChecks.cshtml
```diff
@@ -0,0 +1,195 @@
...
+            </details>
+        </div>
+    </div>
+
+    <script type="text/javascript">
+        document.addEventListener("DOMContentLoaded", function() { 
+            $.get("@Url.Action("OrganisationsWithMultipleActiveAddressesAjax", "AdminDatabaseIntegrityChecks")",
+                "",
+                function(data) { $("#OrganisationsWithMultipleActiveAddresses").html(data)
+                },
```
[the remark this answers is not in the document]
raamSoftwire: Done

GenderPayGap.WebUI/Views/AdminDatabaseIntegrityChecks/DatabaseIntegrityChecks.cshtml
```diff
@@ -0,0 +1,195 @@
...
+                "html"
+              );
+            
+            $.get("@Url.Action("PublicSectorOrganisationsWithoutAPublicSectorTypeAjax", "AdminDatabaseIntegrityChecks")",
+                "",
+                function(data) { $("#PublicSectorOrganisationsWithoutAPublicSectorType").html(data)
+                },
+                "html"
+              );
```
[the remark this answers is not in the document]
raamSoftwire: So it looks like you can't pass js variables into Razor as Razor is server-side and gets converted to html/js by the time you want to pass in your variable. So the way I have written the function for now is 

```
    function loadHtml(url, elementName) {
        $.get("/admin/" + url ,
        "",
        function(data) { $("#" + elementName).html(data)},
        "html"
        );
    }

    document.addEventListener("DOMContentLoaded", function() { 
        loadHtml("active-organisations-with-the-same-name-ajax", "ActiveOrganisationsWithTheSameName");    
    });

```

You can't use `@(Url.Action)` which would have allowed us to use the Action name which is Pascal case rather than kebab-case so that we wouldn't need to enter the same name twice into the function. All of this is a bit sad and I'd be very open to suggestions and improvements.

GenderPayGap.WebUI/Views/AdminDatabaseIntegrityChecks/PublicSectorOrganisationsWithoutAPublicSectorType.cshtml
raamSoftwire: That's neat. I reckon we can do away with the `DatabaseIntegrityChecksViewModel` completely then

GenderPayGap.WebUI/Models/Admin/DatabaseIntegrityChecksViewModel.cs
raamSoftwire: Ahh, I've only just seen this after writing my last comment. Thanks again

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
raamSoftwire: Done

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
raamSoftwire: Thanks

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
raamSoftwire: I think this should solve it ```if (!yearsWithActiveScopes.Contains(year) && !organisationsWithNoActiveScopeForEveryYear.Contains(organisation))
                    {
                        organisationsWithNoActiveScopeForEveryYear.Add(organisation);
                    }```

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
raamSoftwire: Ooo, that's a very good point. I've added that in now

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
raamSoftwire: Done

raamSoftwire:
I'm still waiting for the final OK from Jacob, but I think this is now done

PR #24, opened 2020-02-07
AB#37500 Update view for selecting organisation

raamSoftwire:
Splits up the radio buttons on the select organisation type page. All but the first option mark the organisation as private. 

Previous version

image

New version 

image

### DIFF
--- a/GenderPayGap.WebUI/Views/Register/OrganisationType.cshtml
+++ b/GenderPayGap.WebUI/Views/Register/OrganisationType.cshtml
@@ -1,6 +1,6 @@
 ﻿@using GenderPayGap.Core
-@using GenderPayGap.WebUI.Models.Scope
-@using Microsoft.AspNetCore.Html
+@using GovUkDesignSystem
+@using GovUkDesignSystem.GovUkDesignSystemComponents
 @model GenderPayGap.WebUI.Models.Register.OrganisationViewModel
 @{
     ViewBag.Title = "Your organisation type - Gender pay gap reporting service";
@@ -13,11 +13,11 @@
         {
             @Html.AntiForgeryToken()
 
-            
+
             @Html.HiddenFor(m => m.SearchText)
 
             @await Html.CustomValidationSummaryAsync()
-            
+
             <h1 class="heading-large">
                 Registration options
             </h1>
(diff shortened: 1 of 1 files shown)


GenderPayGap.WebUI/Views/Register/OrganisationType.cshtml
```diff
@@ -31,24 +31,57 @@
...
                 </fieldset>
             </div>
-            <br/>
+            @(Html.GovUkInsetText(new InsetTextViewModel
+            {
+                Html = @<div class="govuk-!-font-size-24">
+                           <p>
+                               If you are not sure what type of organisation you are please read our
+                               <a class="govuk-link" href="https://www.gov.uk/guidance/gender-pay-gap-reporting-overview#relevant-employer">
+                                   guidance</a>.
```
raamSoftwire: If the closing `</a>` tag is on the next line then the link text has an extra unwanted space at the end. Any ideas why? Or how to get round this?

GenderPayGap.WebUI/Views/Register/OrganisationType.cshtml
```diff
@@ -31,24 +31,57 @@
...
+                        <input id="charity" type="radio" name="SectorType" value="@(SectorTypes.Private)">
+                        <label for="charity">
+                            Charity
+                        </label>
+                    </div>
+
                 </fieldset>
             </div>
-            <br/>
+            @(Html.GovUkInsetText(new InsetTextViewModel
```
jamesgriff: Whilst I'm happy for you to use `Html.GovUkInsetText`, the HTML equivalent is very simple, so feel free to use that if you prefer.

I think it's worthwhile using the C# components for any complicated / data-driven components, but for plain / styling components, I'd happy for you to use plain HTML if that makes it clearer.

GenderPayGap.WebUI/Views/Register/OrganisationType.cshtml
jamesgriff: It seems to only do that if you have something after the `</a>` tag (and it doesn't matter whether the something is preceded by whitespace/newline)

raamSoftwire: That's good to know. I'll leave it like this then as we need the full stop

GenderPayGap.WebUI/Views/Register/OrganisationType.cshtml
raamSoftwire: I don't think this view is using the new `GovUkFrontendLayout` so the stylesheets aren't being pulled through. This means that adding `class="govuk-inset-text"` doesn't give us the inset styling. So I used this component as a temporary work-around until we update this page.

raamSoftwire:
This is what it looks like after the copy change. @jamesgriff Jacob has given this the OK, so let me know if you're happy to merge this in now.

image

PR #39, opened 2020-02-18
AB#39439 direct downloads not webjobs

raamSoftwire:
Admin downloads are currently created using a webjob. These changes remove the webjobs and generate the csv files on the fly.

The big benefit is that the download gets the latest data, rather than data gathered up to an hour ago.


GenderPayGap.WebUI/Views/Admin/Home.cshtml
```diff
@@ -76,9 +76,9 @@
 </p>
 
 <p class="govuk-body">
-    <a href="@(Url.Action("Downloads"))" class="govuk-link govuk-!-font-weight-bold">Downloads</a>
+    <a href="@Url.Action("Downloads", "AdminDownloads")" class="govuk-link govuk-!-font-weight-bold">Downloads</a>
     <br/>
-    Download organisation, registration, user, consent.
+    Download user consent.
```
jamesgriff: I changed this line because the first download I switched was the "user consent" download.
Now that we've moved all of them across, we should revert this line, or make it more descriptive.

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
```diff
@@ -0,0 +1,434 @@
...
+
+            var viewModel = new AdminDownloadsViewModel
+            {
+                ReportingYears = Enumerable.Range(firstReportingYear, numberOfYears).Reverse().ToList()
+            };
+
+            return View("Downloads", viewModel);
+        }
+        
+        [HttpGet("downloads-new/all-organisations")]
```
jamesgriff: We can change this to `downloads/all-organisations`

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
```diff
@@ -0,0 +1,434 @@
...
+            var records = userOrganisations.Select(
+                    uo => new
+                    {
+                        uo.OrganisationId,
+                        uo.UserId,
+                        uo.Organisation.OrganisationName,
+                        uo.Organisation.CompanyNumber,
+                        uo.Organisation.SectorType,
+                        uo.Method,
+                        uo.Organisation.LatestScope.ScopeStatus,
```
jamesgriff: Is there always a `LatestScope`.
I could imagine this might be null sometimes, which would cause a NullReferenceException.
I wonder if we even need the scope status - I'd be happy to remove it

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
```diff
@@ -0,0 +1,434 @@
...
```
jamesgriff: Same as above

GenderPayGap.WebUI/Views/Admin/Home.cshtml
raamSoftwire: Good spot. I've now set that to what it was previously

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
```diff
@@ -60,6 +60,35 @@ public FileContentResult DownloadUserConsentSendUpdates()
...
+                        u.ContactFirstName,
+                        u.ContactLastName,
+                        u.ContactJobTitle,
+                        u.ContactEmailAddress,
+                        u.ContactPhoneNumber,
+                        u.ContactOrganisation
+                    })
+                .ToList();
+
+            string fileDownloadName = $"Gpg-UserConsent-AllowContactForFeedback-{DateTime.Now:yyyy-MM-dd HH:mm}.csv";
```
raamSoftwire: Should these be `VirtualDateTime.Now`?

GenderPayGap.WebUI/Controllers/AdminController.cs
```diff
@@ -533,15 +533,6 @@ public async Task<IActionResult> Downloads()
                 model.Downloads.Add(download);
             }
 
-            download = new DownloadViewModel.Download {
-                Type = "Late Submissions",
-                Filepath = Path.Combine(Global.DownloadsPath, Filenames.OrganisationLateSubmissions),
-                Title = "Organisation Late Submissions",
-                Description = "Organisations who reported or changed their submission late the previous snapshot date."
-            };
-            download.ShowUpdateButton = !await GetFileUpdatingAsync(download.Filepath);
```
raamSoftwire: Interesting, I didn't know you could do `!await`

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
raamSoftwire: Oops, missed one. Thanks

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
raamSoftwire: Good point. I've removed this from both userOrganisation downloads

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
jamesgriff: It doesn't really matter - but it's probably best to use `VirtualDateTime` so it's easy to see that we haven't missed any

PR #48, opened 2020-02-26
AB#39493 Fix organisation size column name

raamSoftwire:
Change column name for scope admin download


GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
```diff
@@ -219,7 +219,7 @@ public FileContentResult DownloadAllSubmissionsForYear(int year)
 
                             returnForYear.CompanyLinkToGPGInfo,
                             returnForYear.ResponsiblePerson,
-                            returnForYear.OrganisationSize.GetAttribute<DisplayAttribute>().Name,
```
raamSoftwire: Was there any good reason this was set in the first place?

PR #51, opened 2020-02-27
AB#39631 For user account changes, use AuditLogger instead of logging changes to a file

jamesgriff:


GenderPayGap.BusinessLogic/Repositories/UserRepository.cs
```diff
@@ -214,10 +287,14 @@ public async Task RetireUserAsync(User userToRetire)
             userToRetire.Modified = VirtualDateTime.Now;
 
             // save
-            await DataRepository.SaveChangesAsync();
+            await dataRepository.SaveChangesAsync();
```
raamSoftwire: Here we're still awaiting the `SaveChangesAsync`. But further down we use the more preferred `SaveChangesAsync().Wait()`. Could we make these all work the same way?

raamSoftwire:
Thanks for those changes James, all looks great now

PR #60, opened 2020-03-05
AB#39886 Add visually hidden text for links that opens in a new page

aidayounesi:


GenderPayGap.WebUI/Views/AdminOrganisationSicCodes/ManuallyChangeOrganisationSicCodes.cshtml
```diff
@@ -130,7 +130,7 @@
                 hintOptions: new HintViewModel
                 {
                     Html = @<text>
-                               See <a href="http://resources.companieshouse.gov.uk/sic/" target="_blank" class="govuk-link">Companies House</a> for the official list of SIC codes
+                               See <a href="http://resources.companieshouse.gov.uk/sic/" target="_blank" class="govuk-link">Companies House<span class="visually-hidden">(opens in a new window)</span></a> for the official list of SIC codes
```
jamesgriff: This is one of the new-style pages.
I don't think the `visually-hidden` CSS class is available here.
I'd recommend using `govuk-visually-hidden` instead.

PR #65, opened 2020-03-11
AB#39790 unlabelled textarea

raamSoftwire:
Update TextAreaHtmlGenerator to include `label.for` value

Also updates TextInputHtmlGenerator and CharacterCountHtmlGenerator to make them more consistent


raamSoftwire:
I've done a search for all textareas and all of them have sensible labels.

Other than for the manual changes page, and I think that page has bigger problems/will disappear soon.

But let me know if you think it's worth fixing.

PR #68, opened 2020-03-12
AB#39791 Add some missing labels for input

aidayounesi:


aidayounesi:
Covered: 
<input>
GovUKTextInputFor
CustomEditorFor

GenderPayGap.WebUI/Views/Admin/Home.cshtml
```diff
@@ -22,6 +22,7 @@
 <form method="GET" action="@(Url.Action("SearchGet", "AdminSearch"))">
     <div class="govuk-form-group">
         <label class="govuk-label" for="search-query">
+            <span class="govuk-visually-hidden">Search for</span>
             <b>Organisation:</b> current name, previous name, employer reference, company number
             <br/>
             <b>User:</b> name, email address
```
jamesgriff: This label actually came up when I visited DAC!

There's far too much text in this label.
The screen reader reads out all the test in this label every time the user focusses on this text box!

I think we should split this into:
* A label that says "Search for organisations and users"
* A hint that says the "You can search for organisations by: current name, previous name, employer reference, company number <br> or for users by: name, email address"
https://design-system.service.gov.uk/components/text-input/#hint-text

GenderPayGap.WebUI/Views/Admin/Search.cshtml
```diff
@@ -80,6 +80,7 @@
         <form method="GET" action="@(Url.Action("SearchGet", "AdminSearch"))">
             <div class="govuk-form-group @(Model.Error != null ? "govuk-form-group--error" : "")">
                 <label class="govuk-label" for="search-query">
+                    <span class="govuk-visually-hidden">Search for</span>
```
jamesgriff: See comment in `Admin/Home.cshtml`

jamesgriff:
I made a few changes to the Admin Search hint text.
The main one that's worth noting is the `aria-describedby` on the input that points at the hint.

jamesgriff:
If you're happy with my changes, then feel free to merge this

PR #73, opened 2020-03-13
AB#40151 Use svg search logo

aidayounesi:


GenderPayGap.WebUI/GenderPayGap.WebUI.csproj
```diff
@@ -29,6 +29,7 @@
     <Content Remove="Web.Debug.config" />
     <Content Remove="Web.loadtest.config" />
     <Content Remove="Web.Release.config" />
+    <Content Remove="wwwroot\assets\images\search-button.svg" />
```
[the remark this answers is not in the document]
aidayounesi: It was automatically added. Tested now and removing seems safe.

raamSoftwire:
From my understanding, you can't set the fill if you are using an svg as a background image (https://codepen.io/noahblon/post/coloring-svgs-in-css-background-images). This set me off down a rabbit hole. But it looks like it was actually due to the background position (unless I've got the wrong end of the stick)

raamSoftwire:
image

image

raamSoftwire:
image

image

image

I've submitted a solution that _works_ but there are several reasons I'm not happy with it

- We have to add the svg inline whenever we want to use it
- The inline code needs tweaking in different scenarios (for sizing and position)
- There is a 2px border that isn't quite right on the home page

I'm very open to feedback on this

GenderPayGap.WebUI/wwwroot/styles/components/_admin-search-box.scss
```diff
@@ -4,25 +4,28 @@
...
-        background-image: url('/assets/images/search-button-white-and-black.png');
-        background-repeat: no-repeat;
-        background-position: 2px 50%;
-        text-indent: -5000px;
+        border-style: solid;
+        border-width: 4px 4px 4px 2px;
+        border-color: govuk-colour("black");
         overflow: hidden;
 
+        .svg {
```
jamesgriff: (this is ok for now, but...)
Ideally, we'd either:
* Use a descriptive class name
So, we'd be saying "Make any search icons within the admin-search-button have this style"
e.g.
```
.search-icon {
}
```
* Or, just use the element name
So, we'd be saying "Make any svg elements within the admin-search-button have this style"
e.g.
```
svg {
}

Having an svg who's class is "svg" doesn't add much meaning
```

raamSoftwire: That's a fair point

PR #74, opened 2020-03-16
AB#38814 Admin page to see full history of return details (including all the figures)

jamesgriff:
It's less hideous than last time!

image

### DIFF
--- a/GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationReturnController.cs
+++ b/GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationReturnController.cs
@@ -1,4 +1,5 @@
 ﻿using System.ComponentModel.DataAnnotations;
+using System.IO;
 using System.Linq;
 using GenderPayGap.Core.Interfaces;
 using GenderPayGap.Database;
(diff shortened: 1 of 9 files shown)


GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationReturnController.cs
```diff
@@ -0,0 +1,96 @@
...
+                        LowerMiddleQuarterPercentMale = ret.MaleMiddlePayBand,
+                        LowerMiddleQuarterPercentFemale = ret.FemaleMiddlePayBand,
+                        LowerQuarterPercentMale = ret.MaleLowerPayBand,
+                        LowerQuarterPercentFemale = ret.FemaleLowerPayBand,
+
+                        LinkToCompanyWebsite = ret.CompanyLinkToGPGInfo,
+                        SeniorResponsibleOfficer = ret.ResponsiblePerson
+                    });
+
+            string fileDownloadName = $"ReturnsForOrganisation-{organisation.OrganisationId}--{VirtualDateTime.Now:yyyy-MM-dd HH:mm}.csv";
```
raamSoftwire: [Not important] Would it be more user friendly to use the organisation name here?

jamesgriff: That's a really nice idea :-)

PR #78, opened 2020-03-20
AB#40244 divert private orgs to manual registration

raamSoftwire:


GenderPayGap.WebUI/Views/Register/ChooseOrganisation.cshtml
```diff
@@ -67,7 +67,32 @@
...
+                            
+                            if(!FeatureFlagHelper.IsFeatureEnabled(FeatureFlag.PrivateManualRegistration))
+                            {
+                                <div class="column-two-thirds">
+                                    <p>
+                                        Please choose your organisation and we'll send an activation PIN to its 
+                                        registered address. Only organisations registered at Companies House 
+                                        are listed here.
+                                    </p>
+                                </div>
```
raamSoftwire: This is the old copy and we still want to see it in the case where the feature flag is turned off.

GenderPayGap.WebUI/Views/Register/ChooseOrganisation.cshtml
```diff
@@ -67,7 +67,32 @@
...
+                            }
+                        }
+                        if (FeatureFlagHelper.IsFeatureEnabled(FeatureFlag.PrivateManualRegistration) 
+                            || Model.SectorType == SectorTypes.Public)
```
raamSoftwire: This is the new copy and wants to be shown if

- is a public organisation
- is a private organisation and the flag is on

GenderPayGap.WebUI/Views/Register/RequestReceived.cshtml
```diff
@@ -12,11 +12,11 @@
                     <h1 class="bold-medium">
                         We’ve got your details.
                         <br/>
-                        We will review them and get in touch to confirm.
+                        We will review them and email you to confirm.
```
raamSoftwire: There is no need for conditional logic on this page because this copy change is a general improvement. (And private organisations wouldn't reach this page anyway)

PR #82, opened 2020-03-25
AB#40122 remove user without impersonating

aidayounesi:


GenderPayGap.WebUI/Views/AdminRemoveUser/ConfirmRemoving.cshtml
```diff
@@ -0,0 +1,115 @@
...
+                    },
+                    formGroupOptions: new FormGroupViewModel
+                    {
+                        Classes = "govuk-!-margin-top-6"
+                    })
+
+                @(Html.GovUkButton(new ButtonViewModel
+                {
+                    Text = "Confirm",
+                    Classes = "govuk-!-margin-bottom-4"
```
jamesgriff: This action does an unreversable delete, so I think we should make the button a warning style by adding `govuk-button--warning` to the Classes.

I might also make the Text be something more specific, e.g. `Remove user from organisation`

GenderPayGap.WebUI/Views/AdminRemoveUser/SuccessfullyRemovedUser.cshtml
```diff
@@ -0,0 +1,73 @@
...
+                </a>
+            }
+            else
+            {
+                <a href="@Url.Action("ViewOrganisation", "AdminViewOrganisation", new {id = Model.OrganisationId})"
+                   class="govuk-link">
+                    @(Model.OrganisationName)
+                </a>
+            }
+        </p>
```
jamesgriff: It might be nice to have links to both of these pages

GenderPayGap.WebUI/Views/AdminRemoveUser/SuccessfullyRemovedUser.cshtml
```diff
@@ -0,0 +1,73 @@
...
+    }))
+}
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+        <div class="govuk-panel govuk-panel--confirmation govuk-!-margin-bottom-7">
+            <h1 class="govuk-panel__title govuk-!-font-size-36">
+                User has been removed from the organisation
+            </h1>
```
jamesgriff: When I saw this, I felt like I'd like to be more sure who was removed from which organisation.

Perhaps like this...
image

```cshtml
<h1 class="govuk-panel__title govuk-!-font-size-36">
    @(Model.UserFullName)
    <br/>
    has been removed from
    <br/>
    @(Model.OrganisationName)
</h1>
```

GenderPayGap.WebUI/Controllers/Admin/AdminRemoveUserController.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            viewModel.ParseAndValidateParameters(Request, m => m.Reason);
+
+            if (viewModel.HasAnyErrors())
+            {
+                // If there are any errors, return the user back to the same page to correct the mistakes
+                return View("ConfirmRemoving", viewModel);
+            }
+
+            await RegistrationRepository.RemoveRegistrationAsync(userOrg, user);
```
jamesgriff: (I wouldn't expect you to know this, but...) I think this `RegistrationRepository.RemoveRegistrationAsync` method does:
* more than we need it to do
* and, simultaneously, not everything that we need it to!

It:
* Checks some things for null (which, for us, won't be the case)
* Removes the UserOrg from a list on the Org (but doesn't remove the UserOrg from a similar list on the User (which makes me wonder whether that's necessary at all))
* Changes the Org.LatestRegistration (but there's a case where it might not change it and we'd get an error!)
* Logs some things to a CSV file (which we happen to be moving away from)
* Deletes the UserOrg

I reckon we could reduce this to:
* Change Org.LatestRegistertion to be null (we don't use it for anything important anyway)
* Delete the UserOrg
* Save changes

Then we could remove the dependency on the RegistrationRepository

GenderPayGap.WebUI/Controllers/Admin/AdminRemoveUserController.cs
```diff
@@ -0,0 +1,130 @@
...
+            foreach (string emailAddress in emailAddressesForOrganisation)
+            {
+                EmailSendingService.SendRemovedUserFromOrganisationEmail(
+                    emailAddress,
+                    organisation.OrganisationName,
+                    user.Fullname);
+            }
+
+            // Send the notification to GEO for each newly orphaned organisation
+            if (!user.EmailAddress.StartsWithI(Global.TestPrefix))
```
jamesgriff: We're moving away from using `Global.TestPrefix` (it was only ever used by some old load tests), so we can omit this `if`

GenderPayGap.WebUI/Controllers/Admin/AdminRemoveUserController.cs
```diff
@@ -0,0 +1,130 @@
...
+            {
+                var sendEmails = new List<Task>();
+                bool testEmail = !Config.IsProduction();
+                if (organisation.GetIsOrphan())
+                {
+                    sendEmails.Add(Emails.SendGEOOrphanOrganisationNotificationAsync(organisation.OrganisationName, testEmail));
+                }
+
+                await Task.WhenAll(sendEmails);
```
jamesgriff: I don't think we need the `List<Task>` / `Task.WhenAll()`, since we're only sending one email here.

We could do...
```
if (organisation.GetIsOrphan())
{
    bool testEmail = !Config.IsProduction();
    await Emails.SendGEOOrphanOrganisationNotificationAsync(organisation.OrganisationName, testEmail);
}
```

aidayounesi:
@jamesgriff Thank you for comments. I've addressed all of them.
There is only one point related to a tag and when there are two variables in the url. 
image

PR #85, opened 2020-03-27
AB#40376 Late flag bug fix

raamSoftwire:
Previous PR covered case where a submission was changed after the deadline, but didn't cover case where no submission was present.

### DIFF
--- a/GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
+++ b/GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
@@ -101,18 +101,27 @@
                             {
                                 <tr>
                                     <td colspan="2" style="border-bottom: none;">
-                                        @if (organisation.GetIsInscope(report.AccountingDate.Year))
+                                        @if (Global.ReportingStartYearsToExcludeFromLateFlagEnforcement.Contains(report.AccountingDate.Year) 
+                                             && report.AccountingDate.Year == 2019)
                                         {
-                                            <span>
-                                                This organisation
-                                                @(report.AccountingDate.Year == Global.CurrentAccountingYear ? "is" : "was")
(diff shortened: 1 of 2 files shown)


GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
```diff
@@ -101,18 +101,27 @@
                             {
                                 <tr>
                                     <td colspan="2" style="border-bottom: none;">
-                                        @if (organisation.GetIsInscope(report.AccountingDate.Year))
+                                        @if (Global.ReportingStartYearsToExcludeFromLateFlagEnforcement.Contains(report.AccountingDate.Year)
```
raamSoftwire: This logic means that
- we will only see this message if we are excluding this year
- and if the excluded year is 2019

So if we do turn enforcement back on, then this message will not be shown

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
```diff
@@ -101,18 +101,27 @@
...
-                                        @if (organisation.GetIsInscope(report.AccountingDate.Year))
+                                        @if (Global.ReportingStartYearsToExcludeFromLateFlagEnforcement.Contains(report.AccountingDate.Year) 
+                                             && report.AccountingDate.Year == 2019)
                                         {
-                                            <span>
-                                                This organisation
-                                                @(report.AccountingDate.Year == Global.CurrentAccountingYear ? "is" : "was")
-                                                required to report by @(report.AccountingDate.AddYears(1).AddDays(-1).ToString("d MMMM yyyy"))
-                                            </span>
+                                            <span>Not required to report due to COVID-19</span>
```
raamSoftwire: This is placeholder text. I'm still waiting for feedback from Jacob or Claire here

raamSoftwire:
image

PR #86, opened 2020-03-29
AB#39970 Service desk retire organisation

aidayounesi:


GenderPayGap.WebUI/Views/AdminOrganisationStatus/ViewOrganisationStatus.cshtml
```diff
@@ -0,0 +1,86 @@
...
+{
+    @Html.GovUkButton(new ButtonViewModel
+    {
+        Href = Url.Action("ChangeStatusGet", "AdminOrganisationStatus", new {id = Model.OrganisationId}),
+        Text = "Change status"
+    })
+}
+else
+{
+    <span class="govuk-body-m govuk-!-margin-bottom-4">You cannot change status of an organisation that is @(Model.Status)</span>
```
jamesgriff: It we show this, it might not be clear why we're telling the user that they can't change the status.
It might be nice to put a disabled / unclickable button next to it, to see what they would normally be able to click.
e.g.
```cshtml
<button disabled="disabled" aria-disabled="true" class="govuk-button govuk-button--disabled" data-module="govuk-button" style="vertical-align: baseline;">
    Change status
</button>
<span class="govuk-body" style="color: #626a6e;">
    You cannot change status of an organisation that is @(Model.Status)
</span>
```
**Note:** the `vertical-align: baseline;` lines them up nicely

GenderPayGap.WebUI/Views/AdminOrganisationStatus/ChangeStatus.cshtml
```diff
@@ -0,0 +1,149 @@
...
+                        new FieldsetViewModel
+                        {
+                            Legend = new LegendViewModel
+                            {
+                                Text = "Please select a new status",
+                                Classes = "govuk-fieldset__legend--m"
+                            },
+                            Classes = "govuk-!-margin-top-6"
+                        })
+                }
```
jamesgriff: This might be a nice place to show the explanatory / hint text.
I'm thinking we could use this pattern:
https://design-system.service.gov.uk/components/radios/#radio-items-with-hints

`@Html.GovUkRadioButtonsFor()` doesn't support hints, but we could use the `@Html.GovUkRadios()` method and just provide the details.
This would then mean we had full control over what radios were shown, so we wouldn't need to have 3 separate Enums, which would make things clearer.

What do you think?

GenderPayGap.WebUI/Models/Admin/AdminChangeStatusViewModel.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    }
+
+    public enum NewStatusesFromDeleted
+    {
+
+        Active = 0,
+        Retired = 1
+
+    }
```
jamesgriff: I can see why you've created 3 Enums here, but it feels a little bit surprising / unclear.
I've made a suggestion in `ChangeStatus.cshtml` which might also change how we do this.

GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationStatusController.cs
```diff
@@ -0,0 +1,216 @@
...
+using GenderPayGap.WebUI.Models.Admin;
+using GovUkDesignSystem.Parsers;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace GenderPayGap.WebUI.Controllers
+{
+    [Authorize(Roles = "GPGadmin")]
+    [Route("admin")]
+    public class AdminOrganisationStatusController : Controller
```
jamesgriff: This all looks great 😄

aidayounesi:
@jamesgriff Thank you for the comments. I've addressed them. Also the confirmation page is added.

aidayounesi:
@jamesgriff Thank you for re-reviewing. Are you happy with the code to be merged?

GenderPayGap.WebUI/Views/AdminOrganisationStatus/ChangeStatus.cshtml
```diff
@@ -0,0 +1,173 @@
...
+        },
+        {
+            OrganisationStatuses.Deleted,
+            "An organisation should be Deleted if it was created by mistake / there was no reason for it to ever exist in the service"
+        },
+    };
+
+    var items = new List<ItemViewModel>();
+
+    foreach ((OrganisationStatuses status, string hint) in statusHints)
```
jamesgriff, line 63: 😮 I didn't know this syntax exists - this is so much better than `KeyValuePair`!
This. changes. everything. 😀

PR #87, opened 2020-03-30
AB#40181 fix webjob errors

raamSoftwire:
- Webjob errors in PurgeUser function caused by FK constraint. Unable to delete user because they had audit logs in the DB.
- Updated function deletes the logs first
- I have refactored the function so that if one user fails, the others are still processed
- I have also set up Azure alerts to notify the team DL when functions fail (for all environments)

### DIFF
--- a/GenderPayGap.WebJob/Functions/Functions.Purge.cs
+++ b/GenderPayGap.WebJob/Functions/Functions.Purge.cs
@@ -131,10 +131,7 @@ namespace GenderPayGap.WebJob
             catch (Exception ex)
             {
                 LogFunctionError(runId, nameof(PurgeRegistrations), startTime, ex );
-                
-                //Send Email to GEO reporting errors
-                await _Messenger.SendGeoMessageAsync("GPG - WEBJOBS ERROR", 
-                    $"Failed webjob ({nameof(PurgeRegistrations)}):{ex.Message}:{ex.GetDetailsText()}");
+
                 //Rethrow the error
                 throw;
             }
@@ -233,9 +230,6 @@ namespace GenderPayGap.WebJob
             {
                 LogFunctionError(runId, nameof(PurgeOrganisations), startTime, ex );
                 
-                //Send Email to GEO reporting errors
-                await _Messenger.SendGeoMessageAsync("GPG - WEBJOBS ERROR", 
-                    $"Failed webjob ({nameof(PurgeOrganisations)}):{ex.Message}:{ex.GetDetailsText()}");
                 //Rethrow the error
                 throw;
             }
@@ -265,10 +259,6 @@ namespace GenderPayGap.WebJob
             catch (Exception ex)
             {
                 LogFunctionError(runId, nameof(PurgeTestDataAsync), startTime, ex );
-                
-                //Send Email to GEO reporting errors
-                await _Messenger.SendGeoMessageAsync("GPG - WEBJOBS ERROR", 
-                    $"Failed webjob ({nameof(PurgeTestDataAsync)}):{ex.Message}:{ex.GetDetailsText()}");
 
                 //Rethrow the error
                 throw;
--- a/GenderPayGap.WebJob/Functions/Functions.Purge.cs
+++ b/GenderPayGap.WebJob/Functions/Functions.Purge.cs
@@ -23,7 +23,7 @@ namespace GenderPayGap.WebJob
     {
 
         //Remove any unverified users their addresses, UserOrgs, Org and addresses and archive to zip
-        public async Task PurgeUsers([TimerTrigger("*/5 * * * *" /* every 5 minutes */)] TimerInfo timer, ILogger log)
+        public async Task PurgeUsers([TimerTrigger("40 3 * * *" /* 03:40 once per day */)] TimerInfo timer, ILogger log)
         {
             var runId = CreateRunId();
             var startTime = VirtualDateTime.Now;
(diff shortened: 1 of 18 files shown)


GenderPayGap.WebJob/Functions/Functions.Purge.cs
```diff
@@ -27,48 +28,71 @@ public async Task PurgeUsers([TimerTrigger("40 3 * * *" /* 03:40 once per day */
...
             {
-                DateTime deadline = VirtualDateTime.Now.AddDays(0 - Global.PurgeUnverifiedUserDays);
-                List<User> users = await _DataRepository.GetAll<User>()
-                    .Where(u => u.EmailVerifiedDate == null && (u.EmailVerifySendDate == null || u.EmailVerifySendDate.Value < deadline))
-                    .ToListAsync();
-                DateTime pinExpireyDate = VirtualDateTime.Now.AddDays(0 - Global.PinInPostExpiryDays);
-                foreach (User user in users)
+                //Ignore if they have verified PIN
+                if (user.UserOrganisations.Any(
+                    uo => uo.PINConfirmedDate != null || (uo.PINSentDate != null && uo.PINSentDate < pinExpiryDate)))
```
raamSoftwire: I've run a query against both prod and dev but can't find any results that satisfy this line after having passed line 35. But I am still a bit apprehensive about changing the logic here.

GenderPayGap.WebJob/Functions/Functions.Purge.cs
```diff
@@ -27,48 +28,71 @@ public async Task PurgeUsers([TimerTrigger("40 3 * * *" /* 03:40 once per day */
...
+                    null,
+                    JsonConvert.SerializeObject(new {user.UserId, user.EmailAddress, user.JobTitle, user.Fullname}),
+                    null);
+            
+                DeleteUserAndAuditLogs(user);
+                await Global.ManualChangeLog.WriteAsync(logItem);
             }
             catch (Exception ex)
             {
                 LogFunctionError(runId, nameof(PurgeUsers), startTime, ex );
```
raamSoftwire: This should get picked up by App Insights and the new alert, which means we no longer need to send an email on line 66

GenderPayGap.WebJob/Functions/Functions.Purge.cs
```diff
@@ -27,48 +28,71 @@ public async Task PurgeUsers([TimerTrigger("40 3 * * *" /* 03:40 once per day */
...
+                .Where(log => Equals(log.ImpersonatedUser, user))
+                .Where(log => log.Action == AuditedAction.AdminResendVerificationEmail).ToList();
+            
+            foreach (AuditLog log in auditLogs)
+            {
+                dataRepository.Delete(log);
             }
+            
+            dataRepository.Delete(user);
+            dataRepository.SaveChangesAsync().Wait();
```
raamSoftwire: I haven't been able to test this locally, and I'm not 100% sure this will work. I wonder if I need to try 

`foreach (AuditLog log in auditLogs)
            {
                dataRepository.Delete(log);
            }
            dataRepository.SaveChangesAsync().Wait();
            dataRepository.Delete(user);
            dataRepository.SaveChangesAsync().Wait();`

GenderPayGap.WebJob/Functions/Functions.Purge.cs
```diff
@@ -27,48 +28,71 @@ public async Task PurgeUsers([TimerTrigger("40 3 * * *" /* 03:40 once per day */
...
-                throw;
+        }
+
+        private static void DeleteUserAndAuditLogs(User user)
+        {
+            var dataRepository = Program.ContainerIOC.Resolve<IDataRepository>();
+            
+            List<AuditLog> auditLogs = dataRepository
+                .GetAll<AuditLog>()
+                .Where(log => Equals(log.ImpersonatedUser, user))
```
jamesgriff: I think we should be comparing IDs here rather than the objects themselves.

The objects will be different instances (because they were created by different DataRepositories) so a normal `==` won't work.
You've used `Equals()` which routes through to the `User.Equals()` method (`GenderPayGap.Database.Models\Extensions\User.cs` line 214) which compares IDs.

I think `.Where(log => log.ImpersonatedUser.Id == user.Id)` would be clearest, since we only care about the IDs.

GenderPayGap.WebJob/Functions/Functions.Purge.cs
jamesgriff: Good spot. I'm not sure - it's worth trying

GenderPayGap.WebJob/Functions/Functions.Purge.cs
```diff
@@ -22,53 +23,76 @@ public partial class Functions
     {
 
         //Remove any unverified users their addresses, UserOrgs, Org and addresses and archive to zip
-        public async Task PurgeUsers([TimerTrigger("40 3 * * *" /* 03:40 once per day */)] TimerInfo timer, ILogger log)
+        public async Task PurgeUsers([TimerTrigger("*/5 * * * *" /* every 5 minutes */)] TimerInfo timer, ILogger log)
```
jamesgriff: Just remember to change this back before you merge it

GenderPayGap.WebJob/Functions/Functions.Purge.cs
raamSoftwire: On reflection, I've decided to remove this block. Having run a query on both DEV and PROD I didn't find any users currently in this state. We also know that a user can not add an organisation until after they have been verified so users are unable to get into this state, which means this block should be redundant.

GenderPayGap.WebJob/Functions/Functions.Purge.cs
raamSoftwire: Done

GenderPayGap.WebJob/Functions/Functions.Purge.cs
raamSoftwire: It works the way it is

PR #88, opened 2020-04-01
AB#40561 Add feature flag around PITP journeys

raamSoftwire:
- I've put a feature flag around the block in BaseController. 
- For the `RequestPIN` action on line 262 of RegisterController, it is used in 2 views for requesting a new PIN (ActivateService.cshtml and RequestPIN.cshtml). With the new journey you should be unable to reach either of those pages, but in the `RequestPIN` action I have added an early return if the feature flag is on just in case, which redirects back to the manage organisations page.

----
From the ticket 

RegisterController.cs line 202
The "PINSent" action is the one we want to avoid calling.

If we do a Find Usages, we can see it's referred to in 4 places (+ extras in unit tests)
PINSent.cshtml
"Html.BeginForm()" - this doesn't actually do anything because there's no way to submit the form!

RegisterController.Organisation.cs line 1379
This is correctly guarded by the Feature Flag

RegisterController.cs line 262
Not sure where this is used (we should find out)
This should probably be guarded by the Feature Flag

BaseController.cs line 587
This one is a big risk - this should be guarded by the Feature Flag

### DIFF
--- a/GenderPayGap.WebUI/Classes/BaseClasses/BaseController.cs
+++ b/GenderPayGap.WebUI/Classes/BaseClasses/BaseController.cs
@@ -583,6 +583,11 @@ namespace GenderPayGap.WebUI.Classes
                         {
                             return null;
                         }
+                        
+                        if (FeatureFlagHelper.IsFeatureEnabled(FeatureFlag.PrivateManualRegistration) )
+                        {
+                            return RedirectToAction("ManageOrganisations", "Organisation");
+                        }
 
                         return RedirectToAction("PINSent", "Register");
                     }
--- a/GenderPayGap.WebUI/Controllers/RegisterController.cs
+++ b/GenderPayGap.WebUI/Controllers/RegisterController.cs
@@ -242,6 +242,12 @@ namespace GenderPayGap.WebUI.Controllers
         [HttpPost("request-pin")]
         public async Task<IActionResult> RequestPIN(CompleteViewModel model)
         {
+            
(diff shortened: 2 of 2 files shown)


GenderPayGap.WebUI/Classes/BaseClasses/BaseController.cs
```diff
@@ -572,7 +572,8 @@ protected IActionResult CheckUserRegisteredOk(out User currentUser)
                 return RedirectToAction(nameof(OrganisationController.ManageOrganisations), "Organisation");
             }
 
-            if (userOrg.Organisation.SectorType == SectorTypes.Private)
+            if (userOrg.Organisation.SectorType == SectorTypes.Private 
+                && !FeatureFlagHelper.IsFeatureEnabled(FeatureFlag.PrivateManualRegistration))
```
jamesgriff, line 576: I'm a bit concerned about the placement of this.

This method is called from lots of actions throughout the site, to check if the user is allowed to perform that action.
At the end of this method, there's a `return null` which means "Good to continue with this action".
If we want to prevent the user doing an action (e.g. submitting data for an organisation before their registration is confirmed) then we need to return a `RedirectToAction` somewhere in this method.

I think the change you've made might allow the user to perform actions when they hadn't yet been verified.
e.g. if we have this feature flag turned on, then we omit all the checks between lines 577 and 621.

I think, instead, we should add the feature flag just around line 587 - I'm not sure where we should redirect them to, perhaps the "we have your details and will be in touch" page.

PR #92, opened 2020-04-03
AB#39895 ambiguous buttons

aidayounesi:
I've tried to cover all <input>, <button> and <a> tags

### DIFF
--- a/GenderPayGap.WebUI/Views/AdminFeedback/ViewFeedbackRowsPartial.cshtml
+++ b/GenderPayGap.WebUI/Views/AdminFeedback/ViewFeedbackRowsPartial.cshtml
@@ -43,7 +43,7 @@
                                             data-module="govuk-button"
                                             name="status"
                                             value="@(FeedbackStatus.Spam)">
-                                        Spam
+                                        Spam<span class="govuk-visually-hidden"> feedback created at @(feedback.CreatedDate.ToString("d MMM yyyy HH:mm"))</span>
                                     </button>
                                 }
                                 @if (Model.ShowMarkAsNotSpamButton)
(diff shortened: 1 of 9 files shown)


GenderPayGap.WebUI/Views/AdminFeedback/ViewFeedbackRowsPartial.cshtml
```diff
@@ -43,7 +43,7 @@
                                             data-module="govuk-button"
                                             name="status"
                                             value="@(FeedbackStatus.Spam)">
-                                        Spam
+                                        Spam<span class="govuk-visually-hidden"> feedback created at @(feedback.CreatedDate.ToString("d MMM yyyy HH:mm"))</span>
```
jamesgriff, line 46: I think ideally this would be:
"Mark feedback ID 123 as **Spam**" / "Mark feedback ID 123 as **Not spam**" (where bold means visible)
But then we'd need an invisible entry in the table called Feedback ID

GenderPayGap.WebUI/Views/Register/ChooseOrganisation.cshtml
```diff
@@ -157,15 +161,19 @@
                                                 else
                                                 {
                                                     <li>
-                                                        <button type="submit" class="link-button" name="command" value="page_@i">@i</button>
+                                                        <button type="submit" class="link-button" name="command" value="page_@i">
+                                                            @i<span class="govuk-visually-hidden">th page of organisations list</span>
```
jamesgriff, line 165: "th" isn't always correct (e.g. 2nd)
Maybe "Page i of organisation list" would be better

aidayounesi:
@jamesgriff Thank you for the review. Are you happy with the code to be merged?

PR #93, opened 2020-04-06
AB#40561 Stop unauthorized users reaching manage organisation page

raamSoftwire:


GenderPayGap.WebUI/Controllers/OrganisationController.ManageOrganisations.cs
```diff
@@ -37,7 +37,7 @@ public async Task<IActionResult> ManageOrganisation(string id)
 
             // Check the user has permission for this organisation
             UserOrganisation userOrg = currentUser.UserOrganisations.FirstOrDefault(uo => uo.OrganisationId == organisationId);
-            if (userOrg == null)
+            if (userOrg == null || userOrg.PINConfirmedDate == null)
```
raamSoftwire, line 40: I'm aware that this could be rewritten as `if (userOrg?.PINConfirmedDate == null)` but I think this is clearer as it highlights that we are actually checking for two different states 

- user not yet registered for this organisation
- user registered but not yet verified

jamesgriff: 👍

PR #95, opened 2020-04-06
AB#40691 Remove misleading properties

jamesgriff:
Some of our entities have properties that are misleading or have confusing / imprecise names.
We should replace these with code that more explicitly declares what data it is trying to find.


GenderPayGap.Database.Models/Return.cs
```diff
@@ -49,7 +48,6 @@ public Return()
         public bool EHRCResponse { get; set; }
 
         public virtual Organisation Organisation { get; set; }
```
jamesgriff, line 50: Each `Return` has an `Organisation` - this is expected

GenderPayGap.Database.Models/Return.cs
```diff
@@ -49,7 +48,6 @@ public Return()
         public bool EHRCResponse { get; set; }
 
         public virtual Organisation Organisation { get; set; }
-        public virtual ICollection<Organisation> Organisations { get; set; }
```
jamesgriff, line 52: But each `Return` also has an `Organisations` collection (which is always unused)
This only existed because it mirrors the link from Organisation.LatestReturn to Return (since a Return could, in theory, be linked to by multiple Organisations via their LatestReturn properties)

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
```diff
@@ -96,8 +96,6 @@ public FileContentResult DownloadOrphanOrganisations()
                         Address = org.GetAddressString(),
                         Sector = org.SectorType,
                         ReportingDeadline = org.SectorType.GetAccountingStartDate().AddYears(1).AddDays(-1),
-                        ScopeStatus = org.GetCurrentScope()?.ScopeStatus,
-                        HasSubmitted = org.LatestReturn?.Status == ReturnStatuses.Submitted
```
jamesgriff, line 100: I've realised that these two columns don't make sense out of context.

This is the "Orphan Organisations" download.
This is just a list of the organisations without any users.
It is not being created in the context a particular reporting year.

For each orphan organisation, we give the ScopeStatus (is it in scope) and HasSubmitted (have they submitted).
But, both of these properties only make sense within the context of a particular reporting year.
So, if we give these columns, we're implicitly giving them for a particular reporting year.
This implicit year is different between ScopeStatus and HasSubmitted and can be different between each row - so is useless.

I've removed them for now (they can be joined in Excel if needed).
If we need them in future, we can get more details about the user requirement and build something that works well for them.
But, for the moment, this change is just to stop the data being wrong.

GenderPayGap.WebJob/Functions/Functions.SendReminderEmails.cs
```diff
@@ -89,9 +89,9 @@ private void CheckUserAndSendReminderEmailsForSectorType(User user, SectorTypes
                         || o.LatestScope.ScopeStatus == ScopeStatuses.PresumedInScope))
                 .Where(
                     o =>
-                        o.LatestReturn == null
-                        || o.LatestReturn.AccountingDate != o.SectorType.GetAccountingStartDate()
-                        || o.LatestReturn.Status != ReturnStatuses.Submitted)
+                        !o.Returns.Any(r =>
+                            r.Status == ReturnStatuses.Submitted &&
+                            r.AccountingDate == sector.GetAccountingStartDate()))
```
raamSoftwire, line 94: This is a much nicer articulation of the logic. Thanks James

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
raamSoftwire: I agree and think you have taken the correct approach here

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
```diff
@@ -263,16 +261,24 @@ private static object ConvertReturnToDownloadFormat(Return returnForYear)
...
             List<Organisation> organisationsWithLateReturns = dataRepository.GetAll<Organisation>()
                 .Where(org => org.Status == OrganisationStatuses.Active)
                 .Where(
-                    org => org.LatestScope == null
-                           || org.LatestScope.ScopeStatus == ScopeStatuses.InScope
-                           || org.LatestScope.ScopeStatus == ScopeStatuses.PresumedInScope)
-                .Where(org => org.LatestReturn == null || org.LatestReturn.IsLateSubmission)
+                    org => org.OrganisationScopes.Any(
+                        s => s.SnapshotDate.Year == year
+                             && (s.ScopeStatus == ScopeStatuses.InScope || s.ScopeStatus == ScopeStatuses.PresumedInScope)))
```
raamSoftwire, line 272: Could you explain this one to me a bit further please? Does it still cover the `org.LatestScope == null` case from line 272?

GenderPayGap.Database.Models/Return.cs
raamSoftwire: Thanks for the note

raamSoftwire:
Thanks James, this is a really nice way to simplify the code

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
jamesgriff: Good spot - it doesn't cover that case - I'll fix that

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
```diff
@@ -261,11 +258,8 @@ public async Task<IActionResult> PrivateSectorOrganisationsWithAPublicSectorType
 
             foreach (Organisation organisation in activePrivateOrganisations)
             {
-                bool isInScope =
-                    organisation.LatestScope.Status == ScopeRowStatuses.Active
-                    && (organisation.LatestScope.ScopeStatus == ScopeStatuses.InScope
-                        || organisation.LatestScope.ScopeStatus == ScopeStatuses.PresumedInScope);
-
+                bool isInScope = organisation.GetCurrentScope().ScopeStatus == ScopeStatuses.InScope ||
```
raamSoftwire, line 261: What happens here if the current scope is not active? I assume it returns false, but just want to double check

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
```diff
@@ -77,9 +77,10 @@ public FileContentResult DownloadOrphanOrganisations()
 
             List<Organisation> orphanOrganisations = dataRepository.GetAll<Organisation>()
                 .Where(org => org.Status == OrganisationStatuses.Active)
-                .Where(org => org.LatestScope == null ||
-                              org.LatestScope.ScopeStatus == ScopeStatuses.InScope ||
-                              org.LatestScope.ScopeStatus == ScopeStatuses.PresumedInScope)
+                .Where(org =>
+                    org.OrganisationScopes.OrderByDescending(s => s.SnapshotDate).FirstOrDefault(s => s.Status == ScopeRowStatuses.Active) == null ||
+                    org.OrganisationScopes.OrderByDescending(s => s.SnapshotDate).FirstOrDefault(s => s.Status == ScopeRowStatuses.Active).ScopeStatus == ScopeStatuses.InScope ||
+                    org.OrganisationScopes.OrderByDescending(s => s.SnapshotDate).FirstOrDefault(s => s.Status == ScopeRowStatuses.Active).ScopeStatus == ScopeStatuses.PresumedInScope)
```
raamSoftwire, line 83: These are long lines, is there any change we could break them up?

GenderPayGap.WebUI/Controllers/Admin/AdminDownloadsController.cs
```diff
@@ -109,28 +109,33 @@ public FileContentResult DownloadOrphanOrganisations()
...
-
             List<Organisation> organisationAddresses = dataRepository.GetAll<Organisation>()
                 .Where(org => org.Status == OrganisationStatuses.Active)
-                .Where(org => org.LatestAddress != null)
-                .Include(org => org.LatestAddress)
+                .Include(org => org.OrganisationAddresses)
+                // This .ToList() materialises the collection (i.e. runs the SQL query)
+                .ToList()
+                // We only want organisations with valid addresses
+                // The following filter only works in code (cannot be converted to SQL) so must be done after the first .ToList()
```
raamSoftwire, line 118: That's good to know, thanks for the handy note. Probably worth bringing this to @aidayounesi 's attention too

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
jamesgriff: Good spot - it might throw an error, so I've added a null check

PR #96, opened 2020-04-08
AB#40587 implement create account journey

raamSoftwire:
These changes implement the new design changes.

- content changes
- updating URLs and method names
- tidying up config values

I have also updated the Notify template to cover some copy changes.

I will follow this PR with another one to remove the code that would become obsolete. I think they might need to be released separately.


GenderPayGap.WebUI/Models/Admin/AdminUserResendVerificationEmailController.cs
```diff
@@ -72,21 +74,26 @@ public IActionResult ResendVerificationEmailPost(long id, AdminResendVerificatio
...
+                dataRepository.SaveChangesAsync().Wait();
+            }
+            catch
             {
-                viewModel.AddErrorFor(
-                    m => m.OtherErrorMessagePlaceholder,
-                    "Error whilst re-sending verification email. Please try again in a few minutes.");
-                return View("ResendVerificationEmail", viewModel);
+                // help user resend email
+                throw new Exception("Failed to send verification email. Please try again");
```
jamesgriff, line 96: I don't think the user will see the message of this error.
I think if we throw an exception, they are shown the generic "Something went wrong" message.
In which case, it might be better to **not** catch the exception, so we can see what the original exception is (e.g. some database exception)

GenderPayGap.WebUI/Views/AccountCreation/ResentVerificationCode.cshtml
```diff
@@ -0,0 +1,58 @@
...
+}
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-xl">Confirm your email address</h1>
+    </div>
+</div>
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
```
jamesgriff, line 22: We could probably remove these four lines of code - I don't think there's any value in having two "row"s

GenderPayGap.WebUI/Views/AccountCreation/VerificationCodeExpiredPage.cshtml
```diff
@@ -0,0 +1,53 @@
...
+@using GovUkDesignSystem
+@using GovUkDesignSystem.GovUkDesignSystemComponents
+@model GenderPayGap.Database.User
+@{
+    ViewBag.Title = "Verification code expired - Gender pay gap reporting service";
+    Layout = "~/Views/GovUkFrontend/GovUkFrontendLayout.cshtml";
+}
+
+@if (FeatureFlagHelper.IsFeatureEnabled(FeatureFlag.ReportingStepByStep))
+    {
```
jamesgriff, line 11: (minor) this "{" is oddly-tabbed-in

GenderPayGap.WebUI/Views/AccountCreation/VerificationCodeExpiredPage.cshtml
```diff
@@ -0,0 +1,53 @@
...
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-xl">Your email verification code has expired</h1>
+    </div>
+</div>
+
+
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
```
jamesgriff, line 25: These lines are probably not needed.
Also, the second <div> is oddly-tabbed-in

GenderPayGap.WebUI/Models/Admin/AdminUserResendVerificationEmailController.cs
raamSoftwire: Done. I've now removed the try/catch block

GenderPayGap.WebUI/Views/AccountCreation/ResentVerificationCode.cshtml
raamSoftwire: Done.

raamSoftwire: Actually, I realise why I put in the extra rows. It was so that the step by step sat better, as it would then start in line with the content rather than alongside the title at the top (similar to how it now looks in the guidance). But this can be reviewed later

PR #113, opened 2020-05-05
AB#41078 update empty migrations

raamSoftwire:
Follow up PR to populate empty migrations.

**THIS IS TO BE RELEASED AFTER THE OTHER PR**


GenderPayGap.Database/Migrations/20200505085232_CreateViewsForReportsReader.cs
```diff
@@ -1,12 +1,737 @@
...
+                + " END ");
+
+            migrationBuilder.Sql(" GRANT EXECUTE " + " 	ON OBJECT::[dbo].[OrganisationSectorTypeIdToString] " + " 	TO [ReportsReaderDv]; ");
+
+            #endregion
+
+            #region 201812131250589_Modify view OrganisationInfoView.cs
+
+            migrationBuilder.Sql(
+                " ALTER VIEW [dbo].[OrganisationInfoView] "
```
jamesgriff, line 499: This `ALTER VIEW` replaces the view created on line 66.
I think we can get rid of the block at line 66 and change this block to use the `CREATE VIEW` syntax (which might be as simple as switching `ALTER` for `CREATE`)

GenderPayGap.Database/Migrations/20200505085232_CreateViewsForReportsReader.cs
```diff
@@ -1,12 +1,737 @@
...
+                + " 	FROM dbo.OrganisationPublicSectorTypes AS opst "
+                + " 	INNER JOIN dbo.PublicSectorTypes AS pst ON pst.PublicSectorTypeId = opst.PublicSectorTypeId "
+                + " 		AND opst.Retired IS NULL "
+                + " 	) AS OrgPubSectType ON OrgPubSectType.OrganisationId = orgs.OrganisationId ");
+
+            migrationBuilder.Sql("  GRANT SELECT " + " 	ON OBJECT::dbo.[OrganisationScopeAndReturnInfoView] " + " 	TO [ReportsReaderDv]; ");
+
+            #endregion
+            
+            migrationBuilder.Sql(" ALTER VIEW [dbo].["
```
jamesgriff, line 612: Another `ALTER VIEW` which overrides the view created on line 220.
Similarly, try removing the block and line 220 and changing this block to be a `CREATE`

raamSoftwire:
reopen

PR #116, opened 2020-05-12
AB#41079 convert to postgres

raamSoftwire:
- Install postgres provider for EF Core
- Install Hangfire extensions for postgres
- Change connection strings
- Update any SQL Server specific syntax within EF Core methods


GenderPayGap.Core/Global.cs
```diff
@@ -113,7 +113,7 @@ public static string DownloadsLocation
         public static Version Version => Misc.GetTopAssembly().GetName().Version;
         public static DateTime AssemblyDate => Misc.GetTopAssembly().GetAssemblyCreationTime();
         public static string AssemblyCopyright => Misc.GetTopAssembly().GetAssemblyCopyright();
-        public static string DatabaseConnectionName => Config.GetAppSetting("DatabaseConnectionName") ?? "GpgDatabase";
+        public static string DatabaseConnectionName => Config.GetAppSetting("DatabaseConnectionName") ?? "gpg_database";
```
jamesgriff, line 116: The fact you had to make this change makes me wonder if we're missing this setting in the `appsettings.json` file (maybe actually `appsettings.LOCAL.json`)?

GenderPayGap.Database/GpgDatabaseContext.cs
```diff
@@ -16,7 +16,8 @@ namespace GenderPayGap.Database
     public partial class GpgDatabaseContext : IDbContext
     {
 
-        public static string ConnectionString = @"Server=(localdb)\ProjectsV13;Database=GpgDatabase;Trusted_Connection=True;";
+//        public static string ConnectionString = @"Server=(localdb)\ProjectsV13;Database=GpgDatabase;Trusted_Connection=True;";
+        public static string ConnectionString = @"Server=127.0.0.1;Port=5432;Database=gpg_database;User Id=gpg_user;Password=local_gpg_database;";
```
jamesgriff, line 20: I'd like to be able to switch between using SQL Server and Postgres locally easily.
e.g.
* when you're working on PaaS/DB stuff, you can be using Postgres
* when I'm doing stuff that's going to be deployed to Azure, I can still use SQL Server.

Would it help to make this switch between the two connections strings based on a setting?
Or is that handled somewhere else?

GenderPayGap.Database/GpgDatabaseContextFactory.cs
```diff
@@ -14,7 +14,8 @@ public GpgDatabaseContext CreateDbContext(string[] args)
             var optionsBuilder = new DbContextOptionsBuilder<GpgDatabaseContext>();
 
             //Setup the SQL server with automatic retry on failure
-            optionsBuilder.UseSqlServer(GpgDatabaseContext.ConnectionString, options => options.EnableRetryOnFailure());
+//            optionsBuilder.UseSqlServer(GpgDatabaseContext.ConnectionString, options => options.EnableRetryOnFailure());
+            optionsBuilder.UseNpgsql(@"Server=127.0.0.1;Port=5432;Database=gpg_database;User Id=gpg_user;Password=local_gpg_database;", options => options.EnableRetryOnFailure());
```
jamesgriff, line 18: This probably doesn't need to be hard-coded here if we have the value in `GpgDatabaseContext.ConnectionString`

GenderPayGap.Database/Migrations/20200504111013_InitialCreate.cs
```diff
@@ -873,102 +879,94 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
                 table: "UserStatus",
                 column: "UserId");
-            
-            migrationBuilder.AddForeignKey(
-                name: "FK_OrganisationPublicSectorTypes_Organisations_OrganisationId",
-                table: "OrganisationPublicSectorTypes",
-                column: "OrganisationId",
-                principalTable: "Organisations",
-                principalColumn: "OrganisationId",
-                onDelete: ReferentialAction.Cascade);
```
jamesgriff, line 971: This looks surprising.
I think there should still be a foreign key between `Organisation.OrganisationId` and `OrganisationPublicSectorTypes.OrganisationId`.

I know this is an auto-generated file, so you didn't deliberately make this change.
I can't actually see what would have caused this change to occur!
But I think we should try to make it so this foreign key still exists.

GenderPayGap.WebUI/BackgroundJobs/HangfireConfiguration/HangfireConfigurationHelper.cs
```diff
@@ -13,7 +14,8 @@ public static class HangfireConfigurationHelper
 
         public static void ConfigureServices(IServiceCollection services)
         {
-            services.AddHangfire(x => x.UseSqlServerStorage(Global.DatabaseConnectionString));
+//            services.AddHangfire(x => x.UseSqlServerStorage(Global.DatabaseConnectionString));
+            services.AddHangfire(x => x.UsePostgreSqlStorage(@"Server=127.0.0.1;Port=5432;Database=gpg_database;User Id=gpg_user;Password=local_gpg_database;"));
```
jamesgriff, line 18: I think this should be based on the connection string?
Otherwise it might only work on our laptops (when the database is `127.0.0.1`)

GenderPayGap.WebUI/Startup.cs
```diff
@@ -186,7 +186,7 @@ public IContainer BuildContainerIoC(IServiceCollection services)
             //Register the configuration
             builder.RegisterInstance(Config.Configuration).SingleInstance();
 
-            builder.Register(c => new SqlRepository(new GpgDatabaseContext(Global.DatabaseConnectionString, true)))
+            builder.Register(c => new SqlRepository(new GpgDatabaseContext(@"Server=127.0.0.1;Port=5432;Database=gpg_database;User Id=gpg_user;Password=local_gpg_database;", true)))
```
jamesgriff, line 189: Again, it feels like this should be based on the configurable connection string

GenderPayGap.Core/Global.cs
raamSoftwire: This change has now gone

GenderPayGap.Database/GpgDatabaseContext.cs
raamSoftwire: This is now handled in appsettings.local.json by two values, `UsePostgresDb` and `  "ConnectionStrings": {
    "GpgDatabase": "Server=127.0.0.1;Port=5432;Database=GpgDatabase;User Id=gpg_user;Password=local_gpg_database;"
  }`

raamSoftwire: It is set to false by default in appsettings.json, then overwritten in the local file

GenderPayGap.Database/GpgDatabaseContextFactory.cs
raamSoftwire: All sorted now

GenderPayGap.Database/Migrations/20200504111013_InitialCreate.cs
raamSoftwire: I think I added this manually when I squashed the migrations. It was automatically removed then too, so I had to add it back in to make the schemas match. I'll add it back in again here too.

raamSoftwire:
I have tested switching between SQL Server and Postgres and running the migrations against a fresh db in both cases

PR #117, opened 2020-05-13
AB#41007 Refactor Search

jamesgriff:
* Split out these bits from AdminSearchService
  * SearchRepository
  * SearchCacheUpdaterService
  * 2 cached objects (for Organisations and Users)
* Keep AdminSearchService for the admin-specific search logic
* Make the Admin Search able to search by acronyms (e.g. "bbc" for "British Broadcasting Corporation Limited"

### DIFF
--- a/GenderPayGap.WebUI/Search/CachedObjects/SearchReadyValue.cs
+++ b/GenderPayGap.WebUI/Search/CachedObjects/SearchReadyValue.cs
@@ -1,6 +1,6 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
-using System.Text.RegularExpressions;
 
 namespace GenderPayGap.WebUI.Search.CachedObjects
 {
@@ -17,20 +17,8 @@ namespace GenderPayGap.WebUI.Search.CachedObjects
         public SearchReadyValue(string originalValue)
         {
             OriginalValue = originalValue;
-            LowercaseWords = SplitValueIntoWords(originalValue);
+            LowercaseWords = WordSplittingRegex.SplitValueIntoWords(originalValue);
             Acronym = MakeAcronymFromWords(LowercaseWords);
-        }
-
-        private static List<string> SplitValueIntoWords(string originalValue)
-        {
-            Match match = WordSplittingRegex.Regex.Match(originalValue);
-            if (match.Success)
-            {
-                return match.Groups
-                    .Select(g => g.Value.ToLower())
-                    .ToList();
-            }
-            return new List<string>();
         }
 
         private string MakeAcronymFromWords(List<string> words)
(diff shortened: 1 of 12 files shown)


GenderPayGap.WebUI/Search/CachedObjects/SearchReadyValue.cs
```diff
@@ -0,0 +1,44 @@
...
+        public SearchReadyValue(string originalValue)
+        {
+            OriginalValue = originalValue;
+            LowercaseWords = SplitValueIntoWords(originalValue);
+            Acronym = MakeAcronymFromWords(LowercaseWords);
+        }
+
+        private static List<string> SplitValueIntoWords(string originalValue)
+        {
+            Match match = WordSplittingRegex.Regex.Match(originalValue);
```
raamSoftwire: Will this regex also remove any punctuation (eg. co-op => coop etc)?

GenderPayGap.WebUI/Search/WordSplittingRegex.cs
```diff
@@ -0,0 +1,44 @@
...
+                if (regex != null)
+                {
+                    return regex;
+                }
+
+                // The "language=regexp" comments below tell Resharper to do regex syntax highlighting
+                // "CG" means "character group" - i.e. [A-Z] in regex
+
+                //language=regexp
+                string wordCG = "[a-zA-Z0-9'-]";
```
raamSoftwire: Is a hyphen here counted as a word character then? So `co-op` would remain as `co-op`?

raamSoftwire: It's hard to tell because a hyphen is used for `a-z` too

GenderPayGap.WebUI/Search/SearchRepository.cs
```diff
@@ -19,6 +19,8 @@ public class SearchRepository
         private static List<SearchCachedUser> cachedUsers;
         private static DateTime cacheLastUpdated = DateTime.MinValue;
 
+
+        #region Load data into cache
```
raamSoftwire: I'm still unsure about the use of `#regions`. It is a code smell and one that I was hoping to slowly phase out of the codebase. But I can see that it does a good job here of grouping things together and the file size is still relatively small.

I guess this is just a low-level grumble...

GenderPayGap.WebUI/Search/SearchRepository.cs
```diff
@@ -82,6 +86,8 @@ private List<string> ExtractSearchTermsFromQuery(string query)
                 .ToList();
         }
 
+
+        #region Search Organisations
         private List<AdminSearchResultOrganisationViewModel> SearchOrganisations(string query, List<string> searchTerms)
```
raamSoftwire: The `#regions` do a nice job of providing a narrative, but I think the same is achieved just with the method names here. Which makes the regions feel a bit redundant

GenderPayGap.WebUI/Search/CachedObjects/SearchReadyValue.cs
jamesgriff: Good point, I missed that case.
Just about to push a change that will include this.

GenderPayGap.WebUI/Search/WordSplittingRegex.cs
jamesgriff: Regex patterns are always a bit hard to parse

`[]` means a set of characters that includes `a-z`, `A-Z`, `0-9`, `'` and `-`
So `'` and `-` are part of the word.

The change I'm about to push removes punctuation *after* splitting into words.
So, `co-operative group` becomes [`cooperative`, `group`] rather than [`co`, `operative`, `group`]

PR #119, opened 2020-05-15
AB#41325 test sticky sessions

raamSoftwire:
Remove unused sticky session references


raamSoftwire:
There's one more reference to sticky sessions in the Account Controller in Identity Server

```//Prompt unknown users same as with known users to prevent guessing of valid usernames
                    //Note: Its OK here to use the local web cache since sticky sessions are used by Azure
                    string login = await _cache.GetStringAsync($"{model.Username}:login");```

But the `cache` that it uses is an IDistributedCache, which from my research should be OK? 

https://docs.microsoft.com/en-us/aspnet/core/performance/caching/response?view=aspnetcore-3.1

"**In-memory caching**
In-memory caching uses server memory to store cached data. This type of caching is suitable for a single server or multiple servers using sticky sessions. Sticky sessions means that the requests from a client are always routed to the same server for processing.

**Distributed Cache**
Use a distributed cache to store data in memory when the app is hosted in a cloud or server farm. The cache is shared across the servers that process requests. A client can submit a request that's handled by any server in the group if cached data for the client is available. ASP.NET Core works with SQL Server, Redis, and NCache distributed caches."

jamesgriff:
> There's one more reference to sticky sessions in the Account Controller in Identity Server...

Good spot.
As you said, I think this is just something to test.
Let's see if it works without it.

PR #120, opened 2020-05-15
AB#41015 Wire up homepage auto-complete search to use new SearchRepository

jamesgriff:


GenderPayGap.WebUI/Search/AutoCompleteSearchService.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            query = query.Trim().ToLower();
+
+            List<string> searchTerms = ExtractSearchTermsFromQuery(query);
+
+            List<SearchCachedOrganisation> allOrganisations = SearchRepository.CachedOrganisations;
+
+            List<SearchCachedOrganisation> matchingOrganisations = GetMatchingOrganisations(allOrganisations, searchTerms);
+
+            List<AutoCompleteOrganisation> orderedOrganisations = CalculateRankings(matchingOrganisations, searchTerms, query);
```
raamSoftwire: I assume the `CalculateRankings` also orders the results, based on the variable that you're using. Would it be worth renaming the method name to make this more clear? Something like `OrderOrganisationsByRanking`

GenderPayGap.WebUI/Search/AutoCompleteSearchService.cs
```diff
@@ -0,0 +1,169 @@
...
+        private static List<AutoCompleteOrganisation> CalculateRankings(
+            List<SearchCachedOrganisation> matchingOrganisations,
+            List<string> searchTerms,
+            string query)
+        {
+            return matchingOrganisations
+                .Select(organisation => CalculateRankForOrganisation(organisation, searchTerms, query))
+                .RankHelperOrderByListOfDoubles(name => name.Ranks)
+                .ThenBy(mar => mar.Names[0].Name)
+                .Take(10)
```
raamSoftwire, line 79: From the method name it is also not clear that it was going to truncate that list.

PR #123, opened 2020-05-19
AB#41321 Upgrade to .Net Core 3.1

jamesgriff:


GenderPayGap.WebUI/Startup.cs
```diff
@@ -108,15 +109,20 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
                     })
                 .AddControllersAsServices() // Add controllers as services so attribute filters be resolved in contructors.
                 .SetCompatibilityVersion(CompatibilityVersion.Version_2_2)
-                // Set the default resolver to use Pascalcase instead of the default camelCase which may break Ajaz responses
-                .AddJsonOptions(options => { options.SerializerSettings.ContractResolver = new DefaultContractResolver(); })
-                .AddRazorOptions(
-                    // we need to explicitly set AllowRecompilingViewsOnFileChange because we use a custom environment "Local" for local dev 
-                    // https://docs.microsoft.com/en-us/aspnet/core/mvc/views/view-compilation?view=aspnetcore-2.2#runtime-compilation
-                    options => options.AllowRecompilingViewsOnFileChange = Config.IsLocal())
+                // Set the default resolver to use Pascalcase instead of the default camelCase which may break Ajaz responses
```
raamSoftwire, line 111: Should that be Ajax responses?

GenderPayGap.WebUI/Startup.cs
```diff
@@ -108,15 +109,20 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
...
-                // Set the default resolver to use Pascalcase instead of the default camelCase which may break Ajaz responses
-                .AddJsonOptions(options => { options.SerializerSettings.ContractResolver = new DefaultContractResolver(); })
-                .AddRazorOptions(
-                    // we need to explicitly set AllowRecompilingViewsOnFileChange because we use a custom environment "Local" for local dev 
-                    // https://docs.microsoft.com/en-us/aspnet/core/mvc/views/view-compilation?view=aspnetcore-2.2#runtime-compilation
-                    options => options.AllowRecompilingViewsOnFileChange = Config.IsLocal())
+                // Set the default resolver to use Pascalcase instead of the default camelCase which may break Ajaz responses 
+                .AddJsonOptions(options =>
+                {
+                    options.JsonSerializerOptions.PropertyNameCaseInsensitive = true;
```
raamSoftwire, line 114: This isn't Pascal case. Can you explain what is going on here please?

GenderPayGap.Extensions.AspNetCore/HttpSession.cs
```diff
@@ -91,8 +91,6 @@ public T Get<T>(string key)
                 return default;
             }
 
-            bytes = Encryption.Decompress(bytes);
```
raamSoftwire, line 94: Do we know what the compression was doing? Are we going to need it going forward?

GenderPayGap.WebUI/Startup.cs
jamesgriff: Yes, fixed now

GenderPayGap.WebUI/Startup.cs
jamesgriff: C# property names are typically `PascalCase`
Javascript names are typically `camelCase`

Some of the javascript on our site assumes that properties returned from Ajax calls will be in `PascalCase` (e.g. the homepage auto-complete search)

The ASP.Net JSON library assumes that, when you are serialising objects, you want to serialise to `camelCase` (because that's the normal javascript style).
This would break pages like the auto-complete that assumes `PascalCase`.

So, these lines say "use whatever case the original C# property has" (which is typically `PascalCase`).
I'll make the comment a bit clearer (I copied this block of code from somewhere and din't fix the comment)

GenderPayGap.Extensions.AspNetCore/HttpSession.cs
jamesgriff: I think it was just reducing the size of the session data (which I don't think is important, given that we don't store much in the session and we don't have that many users)

GenderPayGap.WebUI/Startup.cs
raamSoftwire: Thanks for the explanation!

PR #125, opened 2020-05-20
AB#41582 Fix Purge jobs: soft-delete users and organisations

jamesgriff:
The jobs...
* PurgeUsers
* PurgeOrganisations

...were having problems because of Foreign Key constraints between the objects being deleted and the new AuditLog entries that we were adding (to audit the deletion of the object).

This change soft-deletes the objects (so we don't get the Foreign Key conflict).

In future (as part of creating a Data Retention Policy), we should anonymise the records, or delete them completely (but that's for later on)


GenderPayGap.BusinessLogic/Services/AuditLogger.cs
```diff
@@ -108,30 +105,57 @@ private void AuditActionToOrganisation(AuditedAction action, long organisationId
...
+            {
+                impersonatedUserId = session["ImpersonatedUserId"].ToInt64();
+                isImpersonating = impersonatedUserId > 0;
+                originalUserId = session["OriginalUser"].ToInt64();
+            }
+            catch (Exception ex)
+            {
+                // If we are calling the AuditLogger from a background job, then:
+                // - there won't be a session object (so we'll get an Exception)
+                // - the concept of Original and Impersonated users doesn't apply
```
raamSoftwire, line 139: Thanks for this comment

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/PurgeUsersJob.cs
```diff
@@ -27,63 +24,38 @@ public PurgeUsersJob(IDataRepository dataRepository, AuditLogger auditLogger)
...
-                dataRepository.Delete(log);
-            }
+            auditLogger.AuditChangeToUser(
+                AuditedAction.PurgeUser,
+                user,
+                new {user.UserId, user.EmailAddress, user.JobTitle, user.Fullname},
+                null);
 
-            dataRepository.Delete(user);
+            user.SetStatus(UserStatuses.Retired, user, "User retired by PurgeUserJob");
```
raamSoftwire, line 58: Is `retired` for users the same as `deleted` for organisations? Or do they mean different things?

If they mean the same thing, would it be clearer to use the same status phrasing?

jamesgriff: Organisations have both Retired and Deleted:
* Retired is if they've stopped trading, but we still want to be able to see their old returns
* Deleted is if they were registered by mistake (e.g. if someone registered the same organisation twice)

Users just have Retired, which can be for:
* when they close their account
* if they create an account, but don't verify their email address in time

I think it's a good question as to whether Users-Retired is closer to Retired or Deleted, or whether we should have two statuses for the two ways in which they can become retired.
I don't think there's a clear answer either way!

PR #127, opened 2020-05-26
#41623 Remove unused Execute Manual Changes and uses of SecurityCode

jamesgriff:


GenderPayGap.Database.Models/Organisation.cs
```diff
@@ -36,8 +36,11 @@ public Organisation()
         public DateTime? LastCheckedAgainstCompaniesHouse { get; set; }
         public bool OptedOutFromCompaniesHouseUpdate { get; set; } = false;
 
+        [Obsolete]
         public string SecurityCode { get; set; }
+        [Obsolete]
         public DateTime? SecurityCodeExpiryDateTime { get; set; }
+        [Obsolete]
         public DateTime? SecurityCodeCreatedDateTime { get; set; }
```
raamSoftwire, line 44: Why are these marked as obsolete rather than being deleted?

jamesgriff: I'm planning to remove them (and create database migration) in a separate PR

PR #129, opened 2020-05-26
AB#41687 Move all tests into one project

jamesgriff:
Reason: to make the project structure simpler
Bonus benefit: Building the unit tests now takes ~2 minutes rather than ~3.5 minutes :-)


GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/ExtensionsAspNetCoreTests/ConfigTests.cs
```diff
@@ -10,9 +10,9 @@ public class ConfigTests
 
         [Test]
         public void Config_DefaultDate_Returns_DateTime_Now_Plus_Three_Days_And_One_Hour_When_Configured()
```
raamSoftwire, line 12: Should we rename this test to `Config_DefaultDate_Returns_DateTime_Now_Plus_917_Days_When_Configured()`?

PR #131, opened 2020-05-29
AB#41737 Move UserSettings table into User table

jamesgriff:
There is currently a table called UserSettings, which should really be a bunch of columns on the User table.
Storing the data as a separate table causes various problems because the values are stored as strings instead of the real data types (e.g. DateTimes)
This change does that migration.


GenderPayGap.Database.Models/Extensions/User.cs
```diff
@@ -150,58 +120,6 @@ public void SetStatus(UserStatuses status, User byUser, string details = null)
...
-            return verifyUrl;
-        }
-
-        public string GetPasswordResetUrl()
-        {
-            string resetCode = Encryption.EncryptQuerystring(UserId + ":" + VirtualDateTime.Now.ToSmallDateTime());
-            string resetUrl = $"/register/enter-new-password?code={resetCode}";
-            return resetUrl;
-        }
-
```
jamesgriff, line 204: These two methods are just some unrelated unused code

GenderPayGap.Database/GenderPayGap.Database.csproj
```diff
@@ -46,5 +46,10 @@
 
     <!-- Include the layer model as an additional file that will be passed to the analyzer -->
   </ItemGroup>
+  <ItemGroup>
+    <None Update="appsettings.json">
```
raamSoftwire, line 50: Could you explain what is happening with this file please?

GenderPayGap.WebUI/Controllers/Account/AccountCreationController.cs
```diff
@@ -177,8 +177,8 @@ private User CreateNewUser(CreateUserAccountViewModel viewModel)
             user.Lastname = viewModel.LastName;
             user.JobTitle = viewModel.JobTitle;
             user.EmailAddress = viewModel.EmailAddress;
-            user.SetSetting(UserSettingKeys.AllowContact, viewModel.AllowContact.ToString());
-            user.SetSetting(UserSettingKeys.SendUpdates, viewModel.SendUpdates.ToString());
+            user.AllowContact = viewModel.AllowContact;
+            user.SendUpdates = viewModel.SendUpdates;
```
raamSoftwire, line 181: This is so much nicer now!

PR #141, opened 2020-06-12
AB#40974 Change public-facing downloads code to use simplified IFileRepository methods (just Read and Write)

jamesgriff:


GenderPayGap.Core/Classes/Extensions.cs
```diff
@@ -96,15 +96,6 @@ public static DateTime GetAccountingStartDate(this SectorTypes sectorType, int y
...
 
-                //Set the count in the metadata file
-                var count = 0;
-                foreach (object item in records)
-                {
-                    count++;
-                }
-
-                await fileRepository.SetMetaDataAsync(filePath, "RecordCount", count.ToString());
-
```
raamSoftwire, line 107: I don't think we should be making product decisions on the basis that things will be awkward, if these are currently valuable features to the user. Could we capture the features on a ticket in the backlog (if they're not already) so that we can revisit them after the migration to see if there is anything we can do?

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/TestHelpers/NewUiTestHelper.cs
```diff
@@ -65,7 +65,7 @@ public static T GetController<T>(long userId = 0,  Dictionary<string, StringValu
             DIContainer = BuildContainerIoC(dbObjects);
 
             //Create Inversion of Control container
-            MvcApplication.ContainerIoC = DIContainer;
+            Global.ContainerIoC = DIContainer;
```
raamSoftwire, line 68: When you get a chance, could we have a brief chat to discuss this change please? I'd like to know what's going on, the reason behind the change and the implications of moving this IocContainer

GenderPayGap.Core/Classes/Extensions.cs
jamesgriff: Yeh that's fair enough. I've raised a ticket to look at this

PR #143, opened 2020-06-12
#41169 Change Redis config to use VCAP_SERVICES environment variable

jamesgriff:


jamesgriff:
@raamSoftwire I've added a few more changes here - do you mind having another quick look - thanks 😄

PR #145, opened 2020-06-18
AB#41128 Remove unused config settings and simplify config code

jamesgriff:


GenderPayGap.Core/Extensions/AspNetCore/Config.cs
```diff
@@ -71,101 +68,38 @@ public static string EnvironmentName
...
-
-        public static IConfiguration Build(Dictionary<string, string> additionalSettings = null, IConfigurationBuilder builder = null)
+        public static IConfiguration Build(IConfigurationBuilder builder = null)
         {
             builder = builder ?? new ConfigurationBuilder();
 
-            builder.AddJsonFile("appsettings.json", false, true);
-            builder.AddJsonFile($"appsettings.{EnvironmentName}.json", true, true);
-
-            builder.AddEnvironmentVariables();
```
jamesgriff, line 90: This `builder.AddEnvironmentVariables();` is a duplicate - we now just have it once at the end

GenderPayGap.Core/Global.cs
```diff
@@ -131,10 +131,10 @@ public static bool SkipSpamProtection
...
                 {"Referrer-Policy", "origin-when-cross-origin"},
                 {"Strict-Transport-Security", "max-age=31536000; includeSubDomains"},
-                {"X-Powered-By", null},
-                {"X-AspNet-Version", null},
-                {"X-AspNetMvc-Version", null},
-                {"Server", null}
+                {"X-Powered-By", ""},
+                {"X-AspNet-Version", ""},
+                {"X-AspNetMvc-Version", ""},
+                {"Server", ""}
```
raamSoftwire, line 137: Did this work before the change? What did the bug look like?

jamesgriff: This was originally in config (and worked).
Then, I moved it into code, with null values (which didn't work).
The error was one of those Asp.Net errors saying that some middleware threw an exception

PR #148, opened 2020-06-24
AB#41621 Rename Tom's App API controller and add some comments

jamesgriff:


jamesgriff:
There's some talk of having another API (for the Modern Slavery team to connect to).
If we do, I want to have a separate API per client (so that we can change the APIs independently).
If we have a separate API per client, the it makes sense for the API to have a more specific name than just `/api`
And so I wanted to change the URL of this API before we deploy it (other wise it'll be stuck in Tom's App forever and we won't be able to change it!)

GenderPayGap.WebUI/Controllers/ApiForTomsAppController.cs
```diff
@@ -12,13 +12,13 @@
 
 namespace GenderPayGap.WebUI.Controllers
 {
-    [Route("api")]
-    public class ApiController : Controller
+    [Route("api/toms-app")]
```
raamSoftwire, line 15: should we be using the "official" name, the Reports Reader?

jamesgriff: Yeh that makes sense

PR #160, opened 2020-07-20
#41637 Remove views used by Tom's App (and remove the migration that creates them in the first place)

jamesgriff:


jamesgriff:
We probably won't merge / release this in the end, because we'll just turn off the old database instead.

PR #161, opened 2020-07-22
#41102 Add "setup logit-ssl-drain" to Infrastructure-as-code script

jamesgriff:


Infrastructure/CreateEnvironment.sh
```diff
@@ -136,6 +136,16 @@ cf bind-service "gender-pay-gap-${PAAS_ENV_SHORTNAME}" "gpg-${PAAS_ENV_SHORTNAME
 
 
 
+#-------------------
+# Logit.io log drain
+# - Create the log drain service (built by logit.io)
+cf create-user-provided-service logit-ssl-drain -l syslog-tls://0f5c243f-b55b-478f-be38-a7bb80036274-ls.logit.io:18298
```
aidayounesi, line 142: Is it safe to include it in our git repo?

jamesgriff: We're going to ask Logit.io to restrict sending who can send data to this ELK stack to only our PaaS IP addresses, so I think it's ok

PR #162, opened 2020-07-24
GPG-215 Migrate database (changes for PaaS)

jamesgriff:
These are the changes needed to the PaaS instance (to import the data)

### DIFF
--- a/GenderPayGap.Database/Models/AuditLog.cs
+++ b/GenderPayGap.Database/Models/AuditLog.cs
@@ -26,6 +26,7 @@ namespace GenderPayGap.Database.Models
         [JsonProperty]
         public long? ImpersonatedUserId { get; set; }
 
+        [ForeignKey("OrganisationId")]
         public virtual Organisation Organisation { get; set; }
 
         [ForeignKey("OriginalUserId")]
--- a/GenderPayGap.WebUI/Controllers/Admin/AdminDataMigrationController.cs
+++ b/GenderPayGap.WebUI/Controllers/Admin/AdminDataMigrationController.cs
@@ -1,19 +1,19 @@
 using System;
 using System.Collections.Generic;
-using System.ComponentModel.DataAnnotations;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
+using System.Net.Http;
 using System.Text;
-using System.Text.Unicode;
+using System.Threading;
+using Autofac;
 using GenderPayGap.Core;
-using GenderPayGap.Core.Classes;
 using GenderPayGap.Core.Interfaces;
 using GenderPayGap.Database;
 using GenderPayGap.Database.Models;
-using GenderPayGap.Extensions;
 using GenderPayGap.WebUI.Helpers;
 using GenderPayGap.WebUI.Models.Admin;
+using GovUkDesignSystem.Parsers;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
@@ -32,6 +32,7 @@ namespace GenderPayGap.WebUI.Controllers
         {
             this.dataRepository = dataRepository;
         }
+
 
         [HttpGet("data-migration/export-all-as-file-download")]
         public IActionResult ExportAllDataAsFileDownload()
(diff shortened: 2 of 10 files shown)


GenderPayGap.Database/Models/AuditLog.cs
```diff
@@ -1,17 +1,31 @@
...
+        public string DetailsString { get; set; }
+
+        [JsonProperty]
+        public long? OrganisationId { get; set; }
+        [JsonProperty]
+        public long? OriginalUserId { get; set; }
+        [JsonProperty]
+        public long? ImpersonatedUserId { get; set; }
+
         public virtual Organisation Organisation { get; set; }
```
aidayounesi, line 29: For consistency would it be ok to add [ForeignKey("OrganisationId")]

GenderPayGap.WebUI/Controllers/Admin/AdminDataMigrationController.cs
```diff
@@ -0,0 +1,648 @@
...
+            var fileContentResult = new FileContentResult(allDataBytes, "application/json")
+            {
+                FileDownloadName = "AllData.json.gz"
+            };
+
+            return fileContentResult;
+        }
+
+        [AllowAnonymous]
+        [HttpGet("data-migration/export-all")]
```
aidayounesi, line 64: I'm a bit confused about having two HttpGet with the same url

GenderPayGap.WebUI/Controllers/Admin/AdminDataMigrationController.cs
```diff
@@ -0,0 +1,648 @@
...
+        public AdminDataMigrationController(IDataRepository dataRepository)
+        {
+            this.dataRepository = dataRepository;
+        }
+
+
+        [HttpGet("data-migration/export-all")]
+        public IActionResult ExportAllData()
+        {
+            if (dataRepository.GetAll<User>().Count() > 0)
```
aidayounesi, line 40: Why? 😃

GenderPayGap.WebUI/Controllers/Admin/AdminDataMigrationController.cs
```diff
@@ -0,0 +1,648 @@
...
+        {
+            Response.BodyWriter.WriteAsync(Encoding.UTF8.GetBytes(AdminMigrationTopAndBottomOfView.BottomOfView));
+            Response.CompleteAsync().Wait();
+        }
+
+        private void ImportDataFromJsonString(string fileAsString)
+        {
+            AllData allData = JsonConvert.DeserializeObject<AllData>(fileAsString);
+
+            string deleteAllSql = @"
```
aidayounesi, line 283: I reckon we can delete all records of a table by "DELETE FROM TABLE" without any where clause

GenderPayGap.WebUI/Controllers/Admin/AdminDataMigrationController.cs
jamesgriff: Oops, one of these methods was added as a merge error - I'll remove it.
Thanks for spotting this!

GenderPayGap.WebUI/Controllers/Admin/AdminDataMigrationController.cs
jamesgriff: Similarly, this was an earlier version of the code added as a merge error - removed now.
This logic has now been moved into the better-named `IsUserAllowedToImportData`

GenderPayGap.WebUI/Controllers/Admin/AdminDataMigrationController.cs
jamesgriff: I tried to do `DELETE FROM Table` in the UI (DataGrip) but got an error `Unsafe query: 'Delete' statement without 'where' clears all data in the table`.
Maybe this is just in the UI though?

GenderPayGap.Database/Models/AuditLog.cs
jamesgriff: Ah yes, good idea

jamesgriff:
> Question: are we going to remove this code after migration?
I expect we'll remove this code after migration.
Unless we find it very useful for something else (although I'd then want to re-visit it to re-think it's production-worthiness!)

GenderPayGap.WebUI/Controllers/Admin/AdminDataMigrationController.cs
aidayounesi: I'm fairly sure it's just the UI

PR #166, opened 2020-07-29
AB#142 Tables 1 - headers don't refer to data cells

aidayounesi:


GenderPayGap.WebUI/Views/AdminOrganisationSicCodes/ViewOrganisationSicCodes.cshtml
```diff
@@ -46,83 +46,98 @@
...
+    List<OrganisationSicCode> previousSicCodes = Model.OrganisationSicCodes
+        .Where(osc => osc.Retired.HasValue)
+        .OrderByDescending(osc => osc.Retired.Value)
+        .ThenBy(osc => osc.SicCode.SicCodeId)
+        .ToList();
+}
+@if (!previousSicCodes.Any())
+{
+    <div class="govuk-body">
+        No SIC codes from Companies House
```
jamesgriff, line 109: I think this should say `No previous SIC codes`.
This list is SIC codes that the organisation used to have, but were removed.
The SIC codes might or might not be from Companies House - so it's a bit confusing to mention CoHo.

PR #169, opened 2020-08-03
GPG-321 Add HTTP Basic Auth for non-production environments

jamesgriff:
This is just to make the site not publicly accessible / make sure the Googlebot doesn't crawl it - we still have the usual logic system to make sure only authorised users can logic and change data


aidayounesi:
Thank you @jamesgriff looks great

PR #191, opened 2020-08-27
GPG-382 Convert Pages to New Design System

dulciejackson:
### Context
A few of the pages were still using the old design system, so this PR updates those to the newer guidelines found [here](https://design-system.service.gov.uk/)

### Changes
- Updated styling of FeedbackSent.cshtml, ContactUs.cshtml, PrivacyPolicy.cshtml, ReportConcerns.cshtml and Impersonate.cshtml
- Extracted view model and controller for user impersonation from AdminController.cs and updated code style to match newer controllers


jamesgriff (approved):
Hi @dulciejackson,
Looks great - an excellent first PR!
Thanks,
James 😄

PR #192, opened 2020-08-28
GPG-395 Removeability to switch between Postgres and SQL Server (because we always use Postgres now)

jamesgriff:


dulciejackson (approved):
👍

PR #193, opened 2020-08-28
GPG-362 Anonymise Feedback

dulciejackson:
### Context
The new Data Retention Policy means that we should anonymise all PII after 3 years.

> As GEO I want to permanently delete all written feedback on the site three years after submission but keep scoring feedback so that the GPG team can monitor the performance of the site. 

### Changes
- Creates db migration to add HasBeenAnonymised field to Feedback table
- Adds daily job to check for any un-anonymised Feedback which is over 3 years old
- Implements function in job to anonymise any PII to "supplied" and "not supplied" values depending on original data

### DIFF
--- a/GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseThreeYearOldFeedbackJob.cs
+++ b/GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseThreeYearOldFeedbackJob.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using GenderPayGap.Core.Interfaces;
+using GenderPayGap.Database.Models;
 using GenderPayGap.Extensions;
 
 namespace GenderPayGap.WebUI.BackgroundJobs.ScheduledJobs
@@ -17,25 +18,39 @@ namespace GenderPayGap.WebUI.BackgroundJobs.ScheduledJobs
             this.dataRepository = dataRepository;
         }
 
-        public void AnonymiseFeedback()
+        public void AnonymiseFeedbackAction()
         {
             var runId = JobHelpers.CreateRunId();
             var startTime = VirtualDateTime.Now;
-            JobHelpers.LogFunctionStart(runId, nameof(AnonymiseFeedback), startTime);
+            JobHelpers.LogFunctionStart(runId, nameof(AnonymiseFeedbackAction), startTime);
 
             try
             {
-                // TODO: Add clever stuff here
+                DateTime threeYearsAgo = VirtualDateTime.Now.AddYears(-3);
 
-                JobHelpers.LogFunctionEnd(runId, nameof(AnonymiseFeedback), startTime);
+                List<Feedback> feedback = dataRepository.GetAll<Feedback>()
+                    .Where(f => DateTime.Compare(f.CreatedDate, threeYearsAgo) <= 0)
+                    .ToList();
+
+                foreach (Feedback feedbackItem in feedback)
+                {
+                    AnonymiseFeedback(feedbackItem);
+                }
+
+                JobHelpers.LogFunctionEnd(runId, nameof(AnonymiseFeedbackAction), startTime);
             }
             catch (Exception ex)
             {
-                JobHelpers.LogFunctionError(runId, nameof(AnonymiseFeedback), startTime, ex);
+                JobHelpers.LogFunctionError(runId, nameof(AnonymiseFeedbackAction), startTime, ex);
 
                 //Rethrow the error
                 throw;
             }
+        }
+
+        public void AnonymiseFeedback(Feedback feedback)
+        {
+            // TODO: Anonymise individual feedback item here (check what to anonymise to)
         }
 
     }
--- a/GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseThreeYearOldFeedbackJob.cs
+++ b/GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseThreeYearOldFeedbackJob.cs
@@ -30,6 +30,7 @@ namespace GenderPayGap.WebUI.BackgroundJobs.ScheduledJobs
 
                 List<Feedback> feedback = dataRepository.GetAll<Feedback>()
                     .Where(f => DateTime.Compare(f.CreatedDate, threeYearsAgo) <= 0)
+                    .Where(f => !f.HasBeenAnonymised)
                     .ToList();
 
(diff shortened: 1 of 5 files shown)


GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseThreeYearOldFeedbackJob.cs
```diff
@@ -0,0 +1,84 @@
...
+                ? "not supplied"
+                : "supplied";
+
+            feedback.PhoneNumber = string.IsNullOrWhiteSpace(feedback.PhoneNumber)
+                ? "not supplied"
+                : "supplied";
+
+            feedback.Details = string.IsNullOrWhiteSpace(feedback.Details)
+                ? "not supplied"
+                : "supplied";
```
dulciejackson, line 76: I wanted to refactor this into something a bit cleaner, but I didn't want to add unnecessary complexity to something that's not too long and actually pretty readable! I'm not sure if there's a nicer solution than this - I'd love to be able to use JavaScript style `feedback["property name"]` syntax or something similar but [this post](https://stackoverflow.com/questions/10283206/setting-getting-the-class-properties-by-string-name) suggests that it might cause a bunch of overhead that's maybe not worth it to simplify 22 lines of code?

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseThreeYearOldFeedbackJob.cs
```diff
@@ -0,0 +1,84 @@
...
+
+                List<Feedback> feedback = dataRepository.GetAll<Feedback>()
+                    .Where(f => DateTime.Compare(f.CreatedDate, threeYearsAgo) <= 0)
+                    .Where(f => !f.HasBeenAnonymised)
+                    .ToList();
+
+                foreach (Feedback feedbackItem in feedback)
+                {
+                    AnonymiseFeedback(feedbackItem);
+                }
```
jamesgriff, line 39: I think it'd be nice to pull this code out into a separate method - just to make it clear that there's two things going on here:
* code specific to anonymising feedback
* generic code about running web jobs

I can't think of a good name for this method though - sometimes we use something like `AnonymiseFeedbackInner` (but happy for you to think of something better!)

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseThreeYearOldFeedbackJob.cs
```diff
@@ -0,0 +1,84 @@
...
+                ? "not supplied"
+                : "supplied";
+
+            feedback.Details = string.IsNullOrWhiteSpace(feedback.Details)
+                ? "not supplied"
+                : "supplied";
+
+            feedback.HasBeenAnonymised = true;
+
+            dataRepository.SaveChangesAsync().Wait();
```
jamesgriff, line 80: It will be quicker to do the Save outside of the loop (i.e. in the parent method) - that way, we save everything in one go.

There's a slightly irritating thing here about Entity Framework.
EF is very clever and tracks changes to your objects, so it knows what to save.
But, it doesn't know which objects you've finished dealing with.
* The 1st time round this loop, we'd update Feedback 1 and save it - EF would check Feedback 1 and save it
* The 2nd time round this loop, we'd update Feedback 1 and save it - EF would check Feedback 1 and Feedback 2 and save just 2 (but it would have to check both Feedback 1 and 2)

After a while, EF ends up checking loads of objects each time, which gets quite inefficient.
So it's often better to do just one big save at the end.

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseThreeYearOldFeedbackJob.cs
jamesgriff: We could do something like

```csharp
...
feedback.OtherSourceText = Anonymise(feedback.OtherSourceText);
feedback.Details = Anonymise(feedback.Details);
...

private static string Anonymise(string original) {
    return string.IsNullOrWhitespace(original)
        ? "not supplied"
        : "supplied";
}
```

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseThreeYearOldFeedbackJob.cs
```diff
@@ -0,0 +1,84 @@
...
+            var runId = JobHelpers.CreateRunId();
+            var startTime = VirtualDateTime.Now;
+            JobHelpers.LogFunctionStart(runId, nameof(AnonymiseFeedbackAction), startTime);
+
+            try
+            {
+                DateTime threeYearsAgo = VirtualDateTime.Now.AddYears(-3);
+
+                List<Feedback> feedback = dataRepository.GetAll<Feedback>()
+                    .Where(f => DateTime.Compare(f.CreatedDate, threeYearsAgo) <= 0)
```
jamesgriff, line 32: This line can be more simply written as
```csharp
.Where(f => f.CreatedDate < threeYearsAgo)
```

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseThreeYearOldFeedbackJob.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            var runId = JobHelpers.CreateRunId();
+            var startTime = VirtualDateTime.Now;
+            JobHelpers.LogFunctionStart(runId, nameof(AnonymiseFeedbackAction), startTime);
+
+            try
+            {
+                DateTime threeYearsAgo = VirtualDateTime.Now.AddYears(-3);
+
+                List<Feedback> feedback = dataRepository.GetAll<Feedback>()
```
jamesgriff, line 31: Maybe name this variable something like `feedbackItemsToAnonymise`

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseThreeYearOldFeedbackJob.cs
```diff
@@ -0,0 +1,84 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
```
jamesgriff, line 4: It looks like this `using` isn't used

jamesgriff (commented):
A few small changes to make, but otherwise looks great 😄

PR #197, opened 2020-09-02
GPG-396 Create Accessibility Statement Page

dulciejackson:
### Context
We need an accessibility statement in place by 23rd September to be compliant with legislation. An outline of what should be shown on the page can be found [in this Google doc](https://docs.google.com/document/d/1M2LdOu1QgSG9boBDKaE49JaCD-7B4Fdw2Z1EsqsSvBk/edit).

### Changes
- Added 'Accessibility Statement' link to footer, as well as method in HomeController to navigate to it
- Added AccessibilityStatement.cshtml page with content as described in above Google doc

### DIFF
--- a/GenderPayGap.WebUI/Controllers/HomeController.cs
+++ b/GenderPayGap.WebUI/Controllers/HomeController.cs
@@ -85,6 +85,16 @@ namespace GenderPayGap.WebUI.Controllers
         }
 
         #endregion
+        
+        #region Accessibility Statement
+        
+        [HttpGet("~/accessibility-statement")]
+        public IActionResult AccessibilityStatement()
+        {
+            return View("AccessibilityStatement");
+        }
+        
+        #endregion
 
     }
 }
--- a/GenderPayGap.WebUI/Controllers/HomeController.cs
+++ b/GenderPayGap.WebUI/Controllers/HomeController.cs
@@ -22,7 +22,7 @@ namespace GenderPayGap.WebUI.Controllers
         }
 
 
-        [HttpGet("~/ping")]
+        [HttpGet("ping")]
         public IActionResult Ping()
         {
             return new OkResult(); // OK = 200
@@ -38,7 +38,7 @@ namespace GenderPayGap.WebUI.Controllers
 
         #endregion
 
-        [HttpGet("~/report-concerns")]
+        [HttpGet("report-concerns")]
         public IActionResult ReportConcerns()
         {
             return View(nameof(ReportConcerns));
@@ -46,7 +46,7 @@ namespace GenderPayGap.WebUI.Controllers
 
         #region PrivacyPolicy
 
-        [HttpGet("~/privacy-policy")]
+        [HttpGet("privacy-policy")]
         public IActionResult PrivacyPolicy()
         {
             return View("PrivacyPolicy", null);
@@ -54,7 +54,7 @@ namespace GenderPayGap.WebUI.Controllers
 
         [PreventDuplicatePost]
         [ValidateAntiForgeryToken]
-        [HttpPost("~/privacy-policy")]
+        [HttpPost("privacy-policy")]
         public async Task<IActionResult> PrivacyPolicyPost()
         {
             if (!User.Identity.IsAuthenticated)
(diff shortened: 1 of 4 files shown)


GenderPayGap.WebUI/Controllers/HomeController.cs
```diff
@@ -85,6 +85,16 @@ public async Task<IActionResult> PrivacyPolicyPost()
         }
 
         #endregion
+        
+        #region Accessibility Statement
+        
+        [HttpGet("~/accessibility-statement")]
```
jamesgriff, line 91: I think we don't need the `~`

It's only useful if we plan to host the site on a sub-folder (e.g. if we changed from gender-pay-gap.service.gov.uk to gov.uk/gender-pay-gap-service)

We don't plan to do this and it would break loads of other things too, so it's probably best to simplify this to `/accessibility-statement`

GenderPayGap.WebUI/Views/GOVUK/Layouts/template.cshtml
```diff
@@ -191,6 +191,9 @@
                     <li>
                         <a href="/report-concerns">Report Concerns</a>
                     </li>
+                    <li>
+                        <a href="/accessibility-statement">Accessibility Statement</a>
```
jamesgriff, line 195: It would be nicer if this was
```cshtml
<a href="@(Url.Action("AccessibilityStatement", "Home"))">Accessibility statement</a>
```

I know you just copied the ones above, but I prefer the `Url.Action` style because it helps the IDE link the View and the Controler

GenderPayGap.WebUI/Views/Home/AccessibilityStatement.cshtml
```diff
@@ -0,0 +1,122 @@
+﻿@using GenderPayGap.Core
+@{
+    ViewBag.Title = "Accessibility Statement - Gender pay gap reporting service";
```
jamesgriff, line 3: `Gender pay gap service`, not `Gender pay gap reporting service`

GenderPayGap.WebUI/Views/Home/AccessibilityStatement.cshtml
```diff
@@ -0,0 +1,122 @@
...
+            <li>change colours, contrast levels, zoom and fonts</li>
+            <li>complete tasks without a mouse by using input controls such as speech or a keyboard</li>
+            <li>listen using a screen reader</li>
+        </ul>
+        <p class="govuk-body">
+            We’ve also made the website text as simple as possible to understand.
+        </p>
+        <p class="govuk-body">
+            AbilityNet has advice on making your device easier to use if you have a disability.
+        </p>
```
jamesgriff, line 45: In the Google Doc, Tor commented that we might not want this.

Can you check with Tor and either:
* remove this paragraph
OR
* make `AbilityNet` a link to the AbilityNet website

GenderPayGap.WebUI/Views/Home/AccessibilityStatement.cshtml
```diff
@@ -0,0 +1,122 @@
+﻿@using GenderPayGap.Core
+@{
+    ViewBag.Title = "Accessibility Statement - Gender pay gap reporting service";
+    Layout = "~/Views/GovUkFrontend/GovUkFrontendLayout.cshtml";
+}
+
+<div class="govuk-grid-row" xmlns="http://www.w3.org/1999/html">
```
jamesgriff, line 7: I don't think we need the `xmlns` attribute (although let me now if you think we do)

jamesgriff (commented):
A few minor comments, but otherwise excellent.
Thanks 😄

GenderPayGap.WebUI/Views/Home/AccessibilityStatement.cshtml
dulciejackson: I looked at what she was referring to here (she sent me the original GOV.UK version of the accessibility statement) and I think she meant that she wasn't sure whether to make `AbilityNet` a link? On the GOV.UK version it isn't but this line is still in there - happy to clarify this with her! :)

jamesgriff (approved):

PR #199, opened 2020-09-04
GPG-182 Allow Admin to Change Sector

dulciejackson:
### Changes
- Added db table to handle section change history
- Added History/Change link to organisation details page
- Added flow to choose new sector (with descriptions) and confirm with a reason

### Notes
Jacob said that displaying the history here was useful, so I did that as well as handling the change. We agreed for now to remove SIC codes from the organisation when they change sector, but that may be re-visited in future if it becomes problematic.


jamesgriff:
Hi @dulciejackson,

I think the way you're storing sector histories isn't ideal.
This history-style table is useful for data that genuinely has a *history* of changes (e.g. a company changes it's name).
The sector doesn't feel like this - each organisation is only ever in 1 sector, it might just be that we got it wrong to start with!
So this is more a *correction* than a *history*.

It's certainly worth auditing that we made a sector change, but we'll rarely want to see "what sector was X organisation, in Y year" (which we might want to do with names).

I think a better way to do this would be to use the AuditLog.
We currently have a nice way of adding data into the AuditLog.
The "sector history" page could then just read this data back out of the AuditLog table.
If we think of the "sector history" page as the "sector changes" page, then we can just show changes, rather than the full history of values (so for most organisations, this would be an empty list)

If we want to be fancy, we could store entries in the AuditLog using a real class

e.g. instead of:
```csharp
auditLogger.AuditChangeToOrganisation(
    ...,
    new
    {
        OldSector = ...
    });
```

we could write
```csharp
class ChangeSectorAuditLogDetails {
    public SectorTypes OldSector
    ...
}

auditLogger.AuditChangeToOrganisation(
    ...,
    new ChangeSectorAuditLogDetails
    {
        OldSector = ...
    });
```
then we could convert the data straight into the correct class to send to a View

PR #207, opened 2020-09-11
GPG-371 Add Organisation - Manual - Review page

jamesgriff:


GenderPayGap.WebUI/Controllers/AddOrganisation/AddOrganisationManualAddressController.cs
```diff
@@ -46,8 +46,22 @@ public IActionResult ManualAddress(AddOrganisationManualViewModel viewModel)
...
+                {
+                    if (viewModel.Sector.Value == AddOrganisationSector.Private)
+                    {
+                        return RedirectToAction("ManualSicCodes", "AddOrganisationManualSicCodes", viewModel);
+                    }
+                    else
+                    {
+                        return RedirectToAction("ManualReview", "AddOrganisationManualReview", viewModel);
+                    }
+                }
```
dulciejackson, line 64: It might just be me being picky, but this structure of if/else clauses seems a bit weird to me. Both the check for `viewModel.Editing` and the default behaviour for non-private sectors are to redirect to the `ManualReview` function, with one property differing between them.

To me, it'd make more sense to have

```
if(viewModel.Editing == true || viewModel.Sector.Value == AddOrganisationSector.Private) {
    // change viewModel.Editing if needed
    return RedirectToAction("ManualReview", "AddOrganisationManualReview", viewModel);
}
return RedirectToAction("ManualSicCodes", "AddOrganisationManualSicCodes", viewModel);
```

It's essentially the same thing, but saves a bit of duplicate code?

GenderPayGap.WebUI/Controllers/AddOrganisation/AddOrganisationManualChooseSectorController.cs
```diff
@@ -0,0 +1,59 @@
...
+                if (viewModel.Sector.Value == AddOrganisationSector.Private &&
+                    (viewModel.SicCodes == null || viewModel.SicCodes.Count == 0))
+                {
+                    // If the user has selected "Private" and if they haven't selected any SIC codes, then we should prompt them to do this
+                    return RedirectToAction("ManualSicCodes", "AddOrganisationManualSicCodes", viewModel);
+                }
+                else
+                {
+                    return RedirectToAction("ManualReview", "AddOrganisationManualReview", viewModel);
+                }
```
dulciejackson, line 52: This can just be the return line without the surrounding `else`

GenderPayGap.WebUI/Controllers/AddOrganisation/AddOrganisationManualNameController.cs
```diff
@@ -38,8 +38,15 @@ public IActionResult ManualName(AddOrganisationManualViewModel viewModel)
...
-
+                if (viewModel.Editing == true)
+                {
+                    viewModel.Editing = null; // To make the url look a bit nicer (the Review page implies we're editing so there's no need for "Editing" in the url)
+                    return RedirectToAction("ManualReview", "AddOrganisationManualReview", viewModel);
+                }
+                else
+                {
+                    return RedirectToAction("ManualAddress", "AddOrganisationManualAddress", viewModel);
+                }
```
dulciejackson, line 49: This can just be the return line without the surrounding `else`

GenderPayGap.WebUI/Views/AddOrganisationManualChooseSector/ManualChooseSector.cshtml
```diff
@@ -0,0 +1,147 @@
...
+            @(Html.HiddenFor(m => m.County))
+            @(Html.HiddenFor(m => m.Country))
+            @(Html.HiddenFor(m => m.PostCode))
+            @(Html.HiddenFor(m => m.IsUkAddress))
+            @if (Model.SicCodes != null)
+            {
+                @for (int i = 0; i < Model.SicCodes.Count; i++)
+                {
+                    <input type="hidden" name="@(nameof(Model.SicCodes))" value="@(Model.SicCodes[i])" />
+                }
```
dulciejackson, line 49: Might be a bit cleaner to use a foreach with something like 

```
foreach( var sicCode in Model.SicCodes.Select(x,i) => new {Code = x, Index = i){
    <input type="hidden" name="@nameof(sicCode.Code)" value="@(sicCode.Index)" />
}
```

Again this is just personal preference, so I'm not sure whether my suggestion is worth following!

GenderPayGap.WebUI/Views/AddOrganisationManualReview/ManualReview.cshtml
```diff
@@ -0,0 +1,212 @@
...
+            @(Html.HiddenFor(m => m.County))
+            @(Html.HiddenFor(m => m.Country))
+            @(Html.HiddenFor(m => m.PostCode))
+            @(Html.HiddenFor(m => m.IsUkAddress))
+            @if (Model.SicCodes != null)
+            {
+                @for (int i = 0; i < Model.SicCodes.Count; i++)
+                {
+                    <input type="hidden" name="@(nameof(Model.SicCodes))" value="@(Model.SicCodes[i])" />
+                }
```
dulciejackson, line 189: foreach like I suggested before?

dulciejackson (commented):
Looks good, just a couple of suggestions 😄

GenderPayGap.WebUI/Views/AddOrganisationManualChooseSector/ManualChooseSector.cshtml
jamesgriff: I'm not sure why I thought these needed to be for loops.

I've changed these to be:
```cshtml
@foreach (int sicCode in Model.SicCodes)
{
    <input type="hidden" name="@(nameof(Model.SicCodes))" value="@(sicCode)" />
}
```

GenderPayGap.WebUI/Controllers/AddOrganisation/AddOrganisationManualAddressController.cs
jamesgriff: I was trying to communicate why we go to the different pages.
There's 3 distinct cases, so I thought it would be good to have 3 blocks:
* If we're in the "Editing" flow, we go to each page, then back to the "Review" page
* If we're in the "Normal" flow, we go to each page in turn...
  * For private-sector organisations, we need them to add SIC codes, so "SicCodes" is the next page
  * For public-sector organisations, we don't use SIC codes, so the next page is the Review page

2 of the outcomes are the same, but they might well change in future (since we're still deciding on the design of this journey).
I don't want us to assume that these two outcomes are linked, when they're actually quite separate.

It does create quite verbose code, which is a shame.
I've pulled this block (in each of these controllers) out into a separate method, to try to make each method smaller and clearer.
I've also added some comments, to say what each case is for.

GenderPayGap.WebUI/Controllers/AddOrganisation/AddOrganisationManualChooseSectorController.cs
jamesgriff: Personally, I think the `else` makes it clearer, but I think this is probably just a matter of personal taste!

dulciejackson (approved):
Looks good 😄

PR #214, opened 2020-09-17
GPG-182 Allow Admin to Change Organisation Sector

dulciejackson:
Updated branch to fix the git history from GPG-182-allow-admin-to-change-sector!

### DIFF
--- a/GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationSectorController.cs
+++ b/GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationSectorController.cs
@@ -3,6 +3,7 @@ using System.Linq;
 using GenderPayGap.Core;
 using GenderPayGap.Core.Interfaces;
 using GenderPayGap.Database;
+using GenderPayGap.Database.Models;
 using GenderPayGap.Extensions;
 using GenderPayGap.WebUI.Classes;
 using GenderPayGap.WebUI.Helpers;
@@ -33,9 +34,19 @@ namespace GenderPayGap.WebUI.Controllers.Admin
           [HttpGet("organisation/{id}/sector")]
         public IActionResult ViewSectorHistory(long id)
         {
+            var viewModel = new AdminSectorHistoryViewModel();
+            
             var organisation = dataRepository.Get<Organisation>(id);
+            var sectorHistory = dataRepository.GetAll<AuditLog>()
+                .Where(al => al.Action == AuditedAction.AdminChangedOrganisationSector)
+                .Where(al => al.Organisation.OrganisationId == id)
+                .OrderByDescending(al => al.CreatedDate)
+                .ToList();
 
-            return View("ViewOrganisationSector", organisation);
+            viewModel.Organisation = organisation;
+            viewModel.SectorHistory = sectorHistory;
+
+            return View("ViewOrganisationSector", viewModel);
         }
         
         [HttpGet("organisation/{id}/sector/change")]
@@ -194,7 +205,6 @@ namespace GenderPayGap.WebUI.Controllers.Admin
         private void ChangeSector(AdminChangeSectorViewModel viewModel, long organisationId)
         {
             var organisation = dataRepository.Get<Organisation>(organisationId);
-            User currentUser = ControllerHelper.GetGpgUserFromAspNetUser(User, dataRepository);
 
             SectorTypes previousSector = organisation.SectorType;
             SectorTypes newSector = viewModel.NewSector ?? SectorTypes.Unknown;
@@ -211,7 +221,7 @@ namespace GenderPayGap.WebUI.Controllers.Admin
             auditLogger.AuditChangeToOrganisation(
                 AuditedAction.AdminChangedOrganisationSector,
                 organisation,
-                new {PreviousStatus = previousSector, NewStatus = newSector, viewModel.Reason},
+                new AdminChangeSectorAuditLogDetails {OldSector = previousSector, NewSector = newSector, Reason = viewModel.Reason},
                 User);
 
         }
--- a/GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationSectorController.cs
+++ b/GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationSectorController.cs
@@ -68,32 +68,42 @@ namespace GenderPayGap.WebUI.Controllers.Admin
             switch (viewModel.Action)
             {
                 case ChangeOrganisationSectorViewModelActions.OfferNewSectorAndReason:
-                    UpdateAdminChangeSectorViewModelFromOrganisation(viewModel, id);
-                    ValidateAdminChangeSectorViewModel(viewModel);
-                    
-                    // Check if new sector is same as original organisation sector
-                    var newSector = viewModel.NewSector == NewSectorTypes.Private ? SectorTypes.Private : SectorTypes.Public;
-                    if (newSector == viewModel.Organisation.SectorType)
-                    {
-                        viewModel.AddErrorFor(
-                            m => m.NewSector,
-                            "The organisation is already assigned to this sector."
-                        );
-                    }
-                    
-                    if (viewModel.HasAnyErrors())
-                    {
-                        return View("ChangeSector", viewModel);
-                    }
-            
-                    return View("ConfirmSectorChange", viewModel);
+                    return OfferNewSectorAndReason(id, viewModel);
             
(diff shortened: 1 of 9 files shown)


GenderPayGap.WebUI/Views/AdminOrganisationSector/ChangeSector.cshtml
```diff
@@ -0,0 +1,186 @@
...
+                </ul>
+            </text>
+        },
+    };
+
+    var items = new List<ItemViewModel>();
+
+    foreach ((SectorTypes sector, Func<object, object> hint) in sectorHints)
+    {
+        if (Model.Organisation.SectorType != sector)
```
jamesgriff, line 79: As discussed, let's show both (and select the current one)

GenderPayGap.WebUI/Models/Admin/AdminChangeSectorViewModel.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public SectorTypes? NewSector { get; set; }
+
+        [GovUkValidateRequired(ErrorMessageIfMissing = "Please enter a reason for this change.")]
+        public string Reason { get; set; }
+
+        [BindNever /* Output Only - only used for sending data from the Controller to the View */]
+        public Database.Organisation Organisation { get; set; }
+        [BindNever /* Output Only - only used for sending data from the Controller to the View */]
+        public List<InactiveUserOrganisation> InactiveUserOrganisations { get; set; }
```
jamesgriff, line 23: This is never used

GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationSectorController.cs
```diff
@@ -128,6 +183,48 @@ private void AddNewOrganisationPublicSectorType(Organisation organisation, int p
 
             organisation.LatestPublicSectorType = newOrganisationPublicSectorType;
         }
+        
+        private void UpdateAdminChangeSectorViewModelFromOrganisation(AdminChangeSectorViewModel viewModel, long organisationId)
+        {
+            viewModel.Organisation = dataRepository.Get<Organisation>(organisationId);
+
+            viewModel.InactiveUserOrganisations = dataRepository.GetAll<InactiveUserOrganisation>()
+                .Where(m => m.OrganisationId == organisationId).ToList();
```
jamesgriff, line 192: `InactiveUserOrganisations` is never used - this line can be removed

GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationSectorController.cs
```diff
@@ -128,6 +183,48 @@ private void AddNewOrganisationPublicSectorType(Organisation organisation, int p
...
+            viewModel.InactiveUserOrganisations = dataRepository.GetAll<InactiveUserOrganisation>()
+                .Where(m => m.OrganisationId == organisationId).ToList();
+        }
+        
+        private void ValidateAdminChangeSectorViewModel(AdminChangeSectorViewModel viewModel)
+        {
+            if (!viewModel.NewSector.HasValue)
+            {
+                viewModel.AddErrorFor(m => m.NewSector, "Please select a new sector");
+            }
```
jamesgriff, line 200: Instead of this, I think we could use a `GovUkValidateRequired` attribute on the view-model

GenderPayGap.WebUI/Views/AdminOrganisationSector/ChangeSector.cshtml
```diff
@@ -0,0 +1,186 @@
...
+                    </tr>
+                </tbody>
+            </table>
+
+            <div class="govuk-form-group">
+                @{
+                    string error = Model.GetErrorFor<AdminChangeSectorViewModel, SectorTypes?>(m => m.NewSector);
+                    ErrorMessageViewModel radiosErrorMessageViewModel = error != null ? new ErrorMessageViewModel {Text = error} : null;
+                }
+                @Html.GovUkRadios(new RadiosViewModel
```
jamesgriff, line 142: Try using `@Html.GovUkRadiosFor`
See how we do this in `Views/AddOrganisationChooseSector/ChooseSector.cshtml`
I think it might make the code a bit neater (both here and in the controller)

GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationSectorController.cs
```diff
@@ -128,6 +183,48 @@ private void AddNewOrganisationPublicSectorType(Organisation organisation, int p
...
+
+            viewModel.ParseAndValidateParameters(Request, m => m.Reason);
+        }
+        
+        private void ChangeSector(AdminChangeSectorViewModel viewModel, long organisationId)
+        {
+            var organisation = dataRepository.Get<Organisation>(organisationId);
+
+            SectorTypes previousSector = organisation.SectorType;
+            SectorTypes newSector = viewModel.NewSector ?? SectorTypes.Unknown;
```
jamesgriff, line 210: Let's not allow a case where we're setting the sector to Unknown.
If NewSector is null, it's ok to throw an exception - we can even just do NewSector.Value (which will automatically throw an exception if it's null)

GenderPayGap.WebUI/Views/AdminOrganisationSector/ChangeSector.cshtml
```diff
@@ -0,0 +1,168 @@
...
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+        <form method="POST" action="@Url.Action("ChangeSectorPost", "AdminOrganisationSector", new {id = Model.Organisation.OrganisationId})">
+            @(Html.AntiForgeryToken())
+
+            @(Html.Hidden(nameof(AdminChangeSectorViewModel.Action),
+                ChangeOrganisationSectorViewModelActions.OfferNewSectorAndReason))
+
+            @Html.GovUkErrorSummary()
```
jamesgriff, line 61: **(Minor, but useful to know)**

When I tested this, the error messages appear the "wrong" way round.

The fields are in the order (sector, reason) but the errors appear in the order (reason, sector).
The `@Html.GovUkErrorSummary()` has a way of specifying the order of the error messages:
```
@(Html.GovUkErrorSummary(new[]
{
    nameof(AdminChangeSectorViewModel.NewSector),
    nameof(AdminChangeSectorViewModel.Reason)
}))
```

GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationSectorController.cs
```diff
@@ -28,6 +30,71 @@ public class AdminOrganisationSectorController : Controller
...
+                    if (viewModel.HasAnyErrors())
+                    {
+                        return View("ChangeSector", viewModel);
+                    }
+            
+                    return View("ConfirmSectorChange", viewModel);
+            
+                case ChangeOrganisationSectorViewModelActions.ConfirmSectorChange:
+                    ChangeSector(viewModel, id);
+                    return RedirectToAction("ViewSectorHistory", "AdminOrganisationSector", new {id});
```
jamesgriff, line 93: Minor:
I think this code might be a bit easier to read if we move the body of each `case` into a separate method - that way, it's a bit clearer that each case is responding to a different action, rather than it looking like they might both be from the same page

jamesgriff (commented):
A few minor points, but looking good

GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationSectorController.cs
```diff
@@ -128,6 +205,41 @@ private void AddNewOrganisationPublicSectorType(Organisation organisation, int p
...
+            SectorTypes previousSector = organisation.SectorType;
+            SectorTypes newSector = viewModel.NewSector.Value == NewSectorTypes.Private ? SectorTypes.Private : SectorTypes.Public;
+
+            // Update the sector
+            organisation.SectorType = newSector;
+
+            // Remove SIC codes when company changes between sectors
+            organisation.OrganisationSicCodes.Clear();
+
+            dataRepository.SaveChangesAsync().Wait();
```
jamesgriff, line 233: We should also change all the...
* `OrganisationScope.SnapshotDate`
* `Return.AccountingDate`

... to match the new sector.

You can use the method `GetAccountingStartDate` in `GenderPayGap.Core\Classes\Extensions.cs` to get the right value

jamesgriff (commented):
Looks great.
Just one more thing(!):
* We should set the dates to match the new sector type (see comment in file)

jamesgriff (approved):

PR #215, opened 2020-09-18
GPG-409 Confirm in Scope Journey

dulciejackson:
### Context
Replaces the user flow for companies changing from out of scope to in scope.


GenderPayGap.WebUI/Views/ScopeNew/ConfirmInScope.cshtml
```diff
@@ -1,9 +1,107 @@
...
+                new ButtonViewModel
+                {
+                    Text = "Confirm"
+                }
+                ))
+            
+            <p class="govuk-body">
+                <a class="govuk-link" href="@Url.Action("ManageOrganisations", "Organisation")">
+                    Cancel
+                </a>
```
jamesgriff, line 102: I think this *Cancel* link should take the user back to this organisation's page (rather than the organisation list page)
I think we should do this for all the Cancel buttons in these scope journeys

GenderPayGap.WebUI/Views/ScopeNew/ConfirmInScope.cshtml
```diff
@@ -1,9 +1,107 @@
...
+@{
+    string reportingYearFormatted = Model.ReportingYear.Year + "/" + (Model.ReportingYear.Year + 1).ToString().Substring(2, 2);
+    string snapshotDate = new GDSDateFormatter(Model.ReportingYear).FullStartDate;
+}
+
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-        Confirm in scope
+        <form method="POST" action="@Url.Action("ConfirmInScopeAnswers", "ScopeNew", new {encryptedOrganisationId, reportingYear = Model.ReportingYear.Year })">
+            @(Html.AntiForgeryToken())
```
jamesgriff, line 53: This is very minor:
I try to make the `<form>` surround the smallest possible set of elements - to differentiate between:
* data that will be send as part of the form
* data that is just being presented on the page for the user's information

So, this opening `<form>` tag could move down several lines, to just before the `@Html.GovUkButton` (make sure to keep the `@(Html.AntiForgeryToken())` within the form)

GenderPayGap.WebUI/Controllers/ScopeNewController.cs
```diff
@@ -132,6 +130,48 @@ public IActionResult ConfirmOutOfScopeAnswers(string encryptedOrganisationId, in
...
+            if (viewModel.ReportingYear == currentSnapshotDate)
+            {
+                IEnumerable<string> emailAddressesForOrganisation = organisation.UserOrganisations
+                    .Where(uo => uo.PINConfirmedDate.HasValue)
+                    .Select(uo => uo.User.EmailAddress);
+                foreach (string emailAddress in emailAddressesForOrganisation)
+                {
+                    emailSendingService.SendScopeChangeOutEmail(emailAddress, organisation.OrganisationName);
+                }
+            }
```
jamesgriff, line 171: This code is very similar to the Out of Scope journey - it would be good to commonise it.
You could pull the whole lot into a method, or alternatively, pull each component into a method

e.g.
* Retire old scopes for this reporting year
* Add new scope
* Get email addresses for organisation

Have a think about what will be easiest to read/understand.
We're focussing on clarify, not code size, so it would be good to avoid lots of...
```
if (in scope) {
    do X
} else {
    do Y
}
```
...where possible

jamesgriff:
I've just noticed that, in the old code, there's a Confirmation Page for the Out Of Scope journey - I think we should have this in the new code, because it might be the end of the user's interaction with the service, so it's good to be clear that they've completed their action

dulciejackson:
I'll get to commonising the email stuff once I've got a decision from Jacob about when we're wanting to send emails!

jamesgriff (commented):
It looks like there's a problem with the unit tests - the build fails (oddly quickly) for this branch

GenderPayGap.WebUI/Controllers/ScopeNewController.cs
```diff
@@ -132,6 +130,48 @@ public IActionResult ConfirmOutOfScopeAnswers(string encryptedOrganisationId, in
 
             return RedirectToAction("ManageOrganisation", "Organisation", new { id = encryptedOrganisationId });
         }
+        
+        [HttpPost("{encryptedOrganisationId}/reporting-year/{reportingYear}/change-scope/in/confirm")]
```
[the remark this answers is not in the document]
jamesgriff: There's still a couple of `[HttpPost]` actions missing `[ValidateAntiForgeryToken]` attributes

jamesgriff (commented):
Just one small change left now!

jamesgriff (approved):

PR #228, opened 2020-09-25
GPG-182 Adds single sector edge case

dulciejackson:
I tested this on dev and noticed a bug when there's only one sector change, so the Previous sectors header isn't added to the table and the subsequent initial sector line is misaligned:

image

I've fixed this here by adding a check for if the sectorChanges list has count 1 - and if so, I add the Previous sectors header.

I really should have caught this earlier - sorry 😕

### DIFF
--- a/GenderPayGap.WebUI/Views/AdminOrganisationSector/ViewOrganisationSector.cshtml
+++ b/GenderPayGap.WebUI/Views/AdminOrganisationSector/ViewOrganisationSector.cshtml
@@ -123,6 +123,16 @@
                         {
                             // Add a final row for the original scope (not audited so doesn't appear in the list automatically)
                             <tr class="govuk-table__row">
+                                
+                                @if (sectorChanges.Count == 1)
+                                {
+                                    // If this is the only row added (so the Previous sectors header isn't present yet)
+                                    // Add the Previous sectors header
+                                    <th scope="row" class="govuk-table__header" rowspan="@(sectorChanges.Count)">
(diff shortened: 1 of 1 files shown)


GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Administration/AdminOrganisationScopeControllerTests.cs
```diff
@@ -0,0 +1,18 @@
...
+
+        [Test]
+        [Description("POST: Old scopes are retired when a new scope is added")]
+        public void POST_Old_Scopes_Are_Retired_When_New_Scope_Added()
+        {
+            
+        }
+
+    }
+}
```
jamesgriff, line 18: Let's leave this file out unless/until it has some real tests in it

jamesgriff (approved):
Good spot! 😄 
One file to remove, but then happy for you to merge

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Administration/AdminOrganisationScopeControllerTests.cs
dulciejackson: Damn! This has been hanging around in my unstaged files for the other branch where I _actually_ plan to add some tests - didn't mean to commit it!

PR #231, opened 2020-09-28
GPG-410 Scope Change Testing

dulciejackson:
Adds a unit test for scope retirement when a new scope is added, as well as builders for Users, Organisations, and OrganisationScopes.

### DIFF
--- a/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Builders/OrganisationScopeBuilder.cs
+++ b/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Builders/OrganisationScopeBuilder.cs
@@ -8,7 +8,7 @@ namespace GenderPayGap.WebUI.Tests.Builders
     public class OrganisationScopeBuilder
     {
 
-        private Organisation organisation = new OrganisationBuilder();
+        private Organisation organisation = new OrganisationBuilder().Build();
         private bool readGuidance = false;
         private ScopeStatuses scopeStatus = ScopeStatuses.PresumedInScope;
         private string reason = "A reason";
@@ -51,7 +51,7 @@ namespace GenderPayGap.WebUI.Tests.Builders
             return this;
         }
 
-        private OrganisationScope Build()
+        public OrganisationScope Build()
         {
             return new OrganisationScope
             {
@@ -63,11 +63,6 @@ namespace GenderPayGap.WebUI.Tests.Builders
                 Status = status,
                 SnapshotDate = organisation.SectorType.GetAccountingStartDate(reportingYear)
             };
-        }
-        
-        public static implicit operator OrganisationScope(OrganisationScopeBuilder instance)
-        {
-            return instance.Build();
         }
 
     }
--- a/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Builders/OrganisationScopeBuilder.cs
+++ b/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Builders/OrganisationScopeBuilder.cs
@@ -1,0 +1,74 @@
+﻿using System;
+using GenderPayGap.Core;
+using GenderPayGap.Core.Classes;
+using GenderPayGap.Database;
+
+namespace GenderPayGap.WebUI.Tests.Builders
+{
+    public class OrganisationScopeBuilder
+    {
+
+        private Organisation organisation = new OrganisationBuilder();
+        private bool readGuidance = false;
+        private ScopeStatuses scopeStatus = ScopeStatuses.PresumedInScope;
+        private string reason = "A reason";
+        private ScopeRowStatuses status = ScopeRowStatuses.Active;
+        private int reportingYear = 0;
+
+        public OrganisationScopeBuilder WithOrganisation(Organisation organisation)
+        {
+            this.organisation = organisation;
+            return this;
+        }
+
+        public OrganisationScopeBuilder WithReadGuidance(bool readGuidance)
+        {
+            this.readGuidance = readGuidance;
+            return this;
+        }
+
+        public OrganisationScopeBuilder WithScopeStatus(ScopeStatuses scopeStatus)
+        {
+            this.scopeStatus = scopeStatus;
+            return this;
+        }
+
+        public OrganisationScopeBuilder WithReason(string reason)
+        {
+            this.reason = reason;
+            return this;
+        }
+
+        public OrganisationScopeBuilder WithStatus(ScopeRowStatuses status)
+        {
+            this.status = status;
+            return this;
+        }
+
+        public OrganisationScopeBuilder WithReportingYear(int reportingYear)
+        {
+            this.reportingYear = reportingYear;
+            return this;
+        }
+
+        private OrganisationScope Build()
+        {
+            return new OrganisationScope
+            {
+                Organisation = organisation,
+                OrganisationId = organisation.OrganisationId,
+                ReadGuidance = readGuidance,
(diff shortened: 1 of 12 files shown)


GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Builders/OrganisationScopeBuilder.cs
```diff
@@ -0,0 +1,74 @@
...
+using GenderPayGap.Core;
+using GenderPayGap.Core.Classes;
+using GenderPayGap.Database;
+
+namespace GenderPayGap.WebUI.Tests.Builders
+{
+    public class OrganisationScopeBuilder
+    {
+
+        private Organisation organisation = new OrganisationBuilder();
```
jamesgriff, line 11: It took me a while to work out how this could possibly compile!
While the `implicit operator` stuff is very **clever**/**cool**, I don't think it's very **clear**.
Unless there's a good reason, I'd rather we always used `thingBuilder.Build()`

GenderPayGap.WebUI/Controllers/Admin/AdminOrganisationScopeController.cs
```diff
@@ -103,7 +103,7 @@ public async Task<IActionResult> ChangeScopePost(long id, int year, AdminChangeS
             };
 
             dataRepository.Insert(newOrganisationScope);
-            await dataRepository.SaveChangesAsync();
+            dataRepository.SaveChangesAsync().Wait();
```
jamesgriff, line 106: Ooh, if you rebase, you can use the new, synchronous `dataRepository.SaveChanges()` method 😄

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/TestHelpers/NewUiTestHelper.cs
```diff
@@ -209,6 +210,7 @@ public static IContainer BuildContainerIoC(params object[] dbObjects)
             
             //Register all controllers - this is required to ensure KeyFilter is resolved in constructors
             builder.RegisterType<AccountCreationController>().InstancePerLifetimeScope();
+            builder.RegisterType<AdminOrganisationScopeController>().InstancePerLifetimeScope();
```
jamesgriff: I've pushed a snazzy way of doing this that doesn't require us to add a new line for each controller

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Administration/Scope/ScopeControllerTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            OrganisationScope organisationScope2017 = new OrganisationScopeBuilder().WithOrganisation(organisation)
+                .WithReportingYear(2017);
+
+            var requestFormValues = new Dictionary<string, StringValues>();
+            requestFormValues.Add("GovUk_Radio_NewScopeStatus", "OutOfScope");
+            requestFormValues.Add("GovUk_Text_Reason", "A reason");
+
+            object[] dbObjects = {user, organisation, organisationScope2017, organisationScope2018};
+
+            var controller = NewUiTestHelper.GetController<WebUI.Controllers.AdminOrganisationScopeController>(requestFormValues: requestFormValues, dbObjects: dbObjects);
```
jamesgriff: * I've pushed a change to use a new `ControllerBuilder` that (I think) makes this a bit neater
* Are you deliberately  testing the **AdminOrganisationScopeController**?
Your previous work was on the **ScopeController**

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Builders/OrganisationScopeBuilder.cs
dulciejackson: Sure, no worries! I agree that it's clearer - I just looked at a few guides on the builder pattern specific to C# and a handful recommended the implicit operator, so I went with it! Normally I'd always use `.Build()` but I wasn't sure if this was a best practice in C# since a lot of places recommended it!

jamesgriff: Oh that's interesting! Well done for doing the research 😄 

The implicit operator is very clever and does make the code more succinct, which can be nice.
But it can be surprising, and I favour clarify much more than succinctness.

I can why it would be worth it for a lot of organisations - e.g. if we were using loads of builders, and if the organisation did mainly C# so implicit operators were common in the code and wouldn't be surprising.

dulciejackson:
@jamesgriff I've got all the builders setup and working for what I need, but I imagine there's a few things that'll need extending as we test more and find more niche times to use them! Do you think it's ok to leave them here and modify as we go in future? 😄

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Builders/OrganisationScopeBuilder.cs
dulciejackson: Yeah, makes sense. I'm also in favour of `.Build()` for clarity over cool-ness, but I can see how this stuff could be second-nature to C# programmers and end up being recommended for the builder pattern if they're already familiar with it!

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Builders/UserBuilder.cs
```diff
@@ -0,0 +1,79 @@
...
+        public UserBuilder WithStatus(UserStatuses status)
+        {
+            userInProgress.Status = status;
+            return this;
+        }
+
+        public UserBuilder WithOrganisation(Organisation organisation)
+        {
+            UserOrganisation userOrganisation = new UserOrganisationBuilder().ForUser(userInProgress).ForOrganisation(organisation).Build();
+            userInProgress.UserOrganisations.Add(userOrganisation);
```
jamesgriff, line 69: you might also want
```csharp
organisation.UserOrganisations.Add(userOrganisation);
```

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Builders/UserOrganisationBuilder.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using GenderPayGap.Database;
+
+namespace GenderPayGap.WebUI.Tests.Builders
+{
+    public class UserOrganisationBuilder
+    {
+
+        private long userId = 1;
+        private User user = new UserBuilder().Build();
```
jamesgriff, line 10: It would be nice if these two properties were more linked.

e.g. you could remove `userId` and, in the `Build` method, say
```csharp
return new UserOrganisation
{
	UserId = user.UserId, 
	User = user,
	...
};
```

And the same for `organisation` / `organisationId`

jamesgriff (commented):
A couple of minor points, but it's looking really good 😄

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Builders/UserBuilder.cs
dulciejackson: Ah, I didn't know that those would react independently! I guess that makes sense without the clever-ness of EF doing the magic 😄

jamesgriff (approved):

PR #237, opened 2020-10-01
Auto-generate IDs for unit tests (in a thread-safe way)

jamesgriff:
@dulciejackson feel free to use this if you think it's useful - if not, feel free to ignore!


dulciejackson (approved):
That's so useful - definitely better than having to manually set IDs all the time!

PR #238, opened 2020-10-01
GPG-419 Update Change Password Pages

dulciejackson:


GenderPayGap.WebUI/Controllers/Account/ChangePasswordNewController.cs
```diff
@@ -86,6 +86,13 @@ public async void ValidatePasswords(ChangePasswordNewViewModel viewModel, User c
                 viewModel.AddErrorFor(m => m.CurrentPassword, "Could not verify your current password.");
             }
 
+            // Check if new password has 8 or more characters
+            if (viewModel.NewPassword.Length < 8)
+            {
+                viewModel.AddErrorFor(m => m.NewPassword, "New password must be at least 8 characters long");
+            }
```
dulciejackson, line 93: This can be done in the ViewModel with `StringLength`, but I felt like it was more readable to have all this password logic in one place so it's obvious what's being done :)

GenderPayGap.WebUI/Controllers/Account/ChangePasswordNewController.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            // Set up success notification on Manage Account page
+            string nextPageUrl = Url.Action("ManageAccount", "ManageAccount", new {Area = "Account"});
+            StatusMessageHelper.SetStatusMessage(Response, "Your password has been changed successfully", nextPageUrl);
+
+            // Return user to the Manage Account page
+            return LocalRedirect(nextPageUrl);
+        }
+
+        public async void ValidatePasswords(ChangePasswordNewViewModel viewModel, User currentUser)
```
jamesgriff, line 80: Check how we do this in `AccountCreationController` and `CreateUserAccountViewModel` - we use a `[GpgPasswordValidation]` attribute on the Password field which automatically does our password-strength validation for us - so we don't need to do a lot of this manually in the controller

GenderPayGap.WebUI/Models/Account/ChangePasswordNewViewModel.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace GenderPayGap.WebUI.Models.Account
+{
+    public class ChangePasswordNewViewModel : GovUkViewModel
+    {
+
+        [GovUkValidateRequired(ErrorMessageIfMissing = "Please enter your current password")]
+        public string CurrentPassword { get; set; }
+        
+        [GovUkValidateRequired(ErrorMessageIfMissing = "Please enter a new password")]
+        public string NewPassword { get; set; }
```
jamesgriff, line 14: As mentioned, we can use the `[GpgPasswordValidation]` attribute here to do most of our validation

GenderPayGap.WebUI/Controllers/Account/ChangePasswordNewController.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            // Check if current password is correct
+            bool isValidPassword = await userRepository.CheckPasswordAsync(currentUser, viewModel.CurrentPassword);
+            if (!isValidPassword)
+            {
+                viewModel.AddErrorFor(m => m.CurrentPassword, "Could not verify your current password.");
+            }
+
+            // Check if new password has 8 or more characters
+            if (viewModel.NewPassword.Length < 8)
```
jamesgriff, line 90: This caused a NullReferenceException for me because NewPassword wasn't set - we should check that it has a value / it has been parsed successfully before using the value

See how we do this in `AccountCreationController` - e.g.
```csharp
if (viewModel.HasSuccessfullyParsedValueFor(m => m.EmailAddress)
    && viewModel.HasSuccessfullyParsedValueFor(m => m.ConfirmEmailAddress)
    && viewModel.EmailAddress != viewModel.ConfirmEmailAddress)
{
    viewModel.AddErrorFor(
        m => m.ConfirmEmailAddress,
        "The email address and confirmation do not match.");
}
```

GenderPayGap.WebUI/Controllers/Account/ChangePasswordNewController.cs
```diff
@@ -0,0 +1,115 @@
...
+            
+            // Check that passwords are valid
+            ValidatePasswords(viewModel, currentUser);
+
+            if (viewModel.HasAnyErrors())
+            {
+                return View("ChangePasswordNew", viewModel);
+            }
+            
+            userRepository.UpdateUserPasswordUsingPBKDF2(currentUser, viewModel.NewPassword);
```
jamesgriff, line 67: Let's call
```csharp
userRepository.UpdatePassword(currentUser, viewModel.NewPassword);
```
instead of
```csharp
userRepository.UpdateUserPasswordUsingPBKDF2(currentUser, viewModel.NewPassword);
```
because it also audit logs the change

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Account/ChangePasswordController/ChangePasswordControllerTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            controller.AddMockUriHelperNew(new Uri("https://localhost:44371/mockURL").ToString());
+            
+            // Act
+            controller.ChangePasswordPost(new ChangePasswordNewViewModel()).Wait();
+            
+            // Assert
+            bool isExpectedPassword = await mockUserRepo.CheckPasswordAsync(user, "password");
+            Assert.IsTrue(isExpectedPassword);
+        }
+
```
jamesgriff, line 185: We could also have a couple of tests to check:
* Is an audit log saved
* Is an email sent

You an use the new `controllerBuilder.DataRepository` to get newly-created data out of the database (just use it like you would the normal repository)

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Account/ChangePasswordController/ChangePasswordControllerTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            var controller = new ControllerBuilder<ChangePasswordNewController>()
+                .WithUserId(user.UserId)
+                .WithRequestFormValues(requestFormValues)
+                .WithDatabaseObjects(user)
+                .Build();
+            
+            // Required to mock out the Url object when creating the verification URL
+            controller.AddMockUriHelperNew(new Uri("https://localhost:44371/mockURL").ToString());
+
```
jamesgriff, line 119: It might be nice to add this to the controller builder
It looks like we don't care what the value of the URL is, so it would be great if we could avoid having `"https://localhost:44371/mockURL"` in this file (because it makes you think it's important to the test)

GenderPayGap.WebUI/Controllers/Account/ChangePasswordNewController.cs
```diff
@@ -0,0 +1,115 @@
...
+using GenderPayGap.WebUI.Repositories;
+using GenderPayGap.WebUI.Services;
+using GovUkDesignSystem;
+using GovUkDesignSystem.Parsers;
+using Microsoft.AspNetCore.Mvc;
+
+namespace GenderPayGap.WebUI.Controllers.Account
+{
+    [Route("manage-account")]
+    public class ChangePasswordNewController : Controller
```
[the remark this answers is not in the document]
dulciejackson: Can only logged-in employers change their password then? I'm a bit confused about how the roles and who fits under which one - do all people reporting on behalf of an organisation fall under the GpgEmployer role?

GenderPayGap.WebUI/Models/Account/ChangePasswordNewViewModel.cs
dulciejackson: Does this cover our basic password expectations - one upper case, one lower case, one number and >8 characters? I hadn't seen this - very clever!

GenderPayGap.WebUI/Controllers/Account/ChangePasswordNewController.cs
dulciejackson: Is there a reason `UpdateUserPasswordUsingPBKDF2` is public since we're likely to always want to audit log the change? I picked the wrong one here because I didn't realise that one was called within the other!

jamesgriff: Good point - no, I think we could:
* change the only usage of it from outside the class to use `UpdatePassword`
* remove `UpdateUserPasswordUsingPBKDF2` from the interface
* make `UpdateUserPasswordUsingPBKDF2` private

GenderPayGap.WebUI/Controllers/Account/CloseAccountNewController.cs
```diff
@@ -14,11 +12,13 @@
 using GenderPayGap.WebUI.Services;
 using GovUkDesignSystem;
 using GovUkDesignSystem.Parsers;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 
 namespace GenderPayGap.WebUI.Controllers.Account
 {
     [Route("manage-account")]
+    [Authorize(Roles = LoginRoles.GpgEmployer)]
```
jamesgriff, line 21: This line is great.
But, I've just realised that it might break the `close/completed` view (because the user will be logged out at that stage).
We should test that (and add `[AllowAnonymous]` on the `close/completed` action if needed)

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Account/ChangePasswordController/ChangePasswordControllerTests.cs
```diff
@@ -0,0 +1,245 @@
...
+            controller.ChangePasswordPost(new ChangePasswordNewViewModel()).Wait();
+            
+            // Assert
+            // Assert that exactly one email is sent
+            Assert.That(controllerBuilder.EmailsSent.Count == 1);
+            
+            NotifyEmail userEmail = controllerBuilder.EmailsSent.FirstOrDefault();
+            // Assert that the email sent has the correct email address and template
+            Assert.NotNull(userEmail);
+            Assert.AreEqual(userEmail.EmailAddress, user.EmailAddress);
```
jamesgriff, line 100: A couple of really minor points about how you're using the Assertion library:

If we use it correctly, Assert.* can give us reasonably helpful error messages, e.g.
```
Test Failure : TestClassName.TestMethodName
  Expected: user1@example.com
  But was:  user2@example.com
```

To make this work, we need to:
* Use `Assert.AreEqual()` rather than `Assert.That()`
* Make sure we get the parameters the right way round (expected result first, then actual result) - otherwise the message will look confusing!

In this case, the parameters are the wrong way round

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Account/ChangePasswordController/ChangePasswordControllerTests.cs
```diff
@@ -0,0 +1,245 @@
...
+            // Act
+            controller.ChangePasswordPost(new ChangePasswordNewViewModel()).Wait();
+            
+            // Assert
+            // Assert that exactly one audit log entry is added
+            List<AuditLog> auditLogEntries = controllerBuilder.DataRepository.GetAll<AuditLog>().ToList();
+            Assert.That(auditLogEntries.Count == 1);
+            
+            // Assert that the audit log entry audits the correct action
+            Assert.That(auditLogEntries.First().Action == AuditedAction.UserChangePassword);
```
jamesgriff, line 133: And here you could use `Assert.AreEqual` instead of `Assert.That`

jamesgriff (commented):
A couple of really minor points!

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Account/ChangePasswordController/ChangePasswordControllerTests.cs
dulciejackson: Ah ok, I didn't realise that the choice made a difference to the error message, but that makes sense now I think about it!

jamesgriff (approved):
Looks great - nice one :-D

PR #240, opened 2020-10-02
GPG-422 Update Close Account Pages

dulciejackson:


GenderPayGap.WebUI/Controllers/Account/CloseAccountNewController.cs
```diff
@@ -91,17 +91,19 @@ public async Task<IActionResult> CloseAccountPost(CloseAccountNewViewModel viewM
...
             }
 
-            // Collect list of organisations that will be orphaned once the user's account is closed
-            List<Organisation> orphanedOrganisations = currentUser.UserOrganisations
+            // Collect list of organisations associated with the user
+            List<Organisation> userOrganisations = currentUser.UserOrganisations
                 .Select(uo => uo.Organisation)
                 .Distinct()
-                .Where(org => org.UserOrganisations.Count == 1)
                 .ToList();
```
dulciejackson: Before retiring the user account (which includes removing all UserOrganisations), collect a list of Organisations associated with the user. We _could_ infer that any Organisation with only one UserOrganisation associated will become an orphan, but this doesn't feel very readable - so instead I'm turning the list of all Organisations into a sublist of just orphaned ones on line 105 - after the UserOrganisations associated with this User have been removed.

jamesgriff (approved):

PR #245, opened 2020-10-07
GPG-504 Turn off Hangfire retries in non-production environments (to reduce alert-spam to developers)

jamesgriff:


dulciejackson (approved):
I'm slightly disappointed - who doesn't love starting a Monday morning by deleting 200 failure emails? 😄 🎉

PR #250, opened 2020-10-09
GPG-419 Remove Old Change Password Code

dulciejackson:


GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Account/ChangePasswordController/ChangePasswordControllerTests.cs
```diff
@@ -52,15 +51,15 @@ public async Task POST_Password_Is_Updated_As_Expected()
             requestFormValues.Add("GovUk_Text_NewPassword", "NewPassword1");
             requestFormValues.Add("GovUk_Text_ConfirmNewPassword", "NewPassword1");
 
-            var controller = new ControllerBuilder<ChangePasswordNewController>()
+            var controller = new ControllerBuilder<WebUI.Controllers.Account.ChangePasswordController>()
```
jamesgriff, line 54: Just so you know, the only reason we need the `WebUI.Controllers.Account.` is because this file is inside a folder called `ChangePasswordController`.

If we put this file in the parent folder (which would be a reasonable place to put it) then we wouldn't need to specify this (we could just use a `using` statement instead)

jamesgriff (approved):
One minor (optional) suggestion, but otherwise happy for you to merge

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Account/ChangePasswordController/ChangePasswordControllerTests.cs
dulciejackson: Ah ok, that's good to know. When I created these I just assumed that they'd be best inside the pre-existing Account folder, but I'm happy to change that if you'd rather they weren't in a sub-folder!

jamesgriff: Typically, test files live parallel to the file they're testing, so this would be in the Account folder.

If we had loads of tests for a particular file, then it would make sense to split the test file up into smaller files and create a folder with the name of the file being tested.
That's the pattern we currently have here, but with only 1 file in the sub-folder.

That's a sensible general pattern to follow, but I don't mind if we deviate from it now and then!

Binney (approved):

PR #253, opened 2020-10-13
GPG-423 Update Manage Account Page

dulciejackson:
### Context
All the links from this page have been updated to the new design system, so now this is being updated to the new design system (at last)!

### Notes
After this is all reviewed I'll create a second PR to swap the links over to pointing at this page instead of the old version 😄


jamesgriff:
@dulciejackson 
I went on a bit of a coding spree and made lots of little improvements.
It generally looked good - I just though I'd try to simplify a bunch of things.
Can you review my changes and see if you agree with them.
Thanks :-)

jamesgriff (approved):

PR #255, opened 2020-10-14
GPG-510 Removes old Manage Account code

dulciejackson:
Removes the old Manage Account code - the folders are still there because ManageOrganisation still uses the old AccountNavigation component (the only thing still in them), but the Areas folder will be gone soon! 🎉 

I removed the old tests but it didn't seem necessary to write new ones - they'd basically just be testing that we've got the right auth checks and annotations on the GET!


jamesgriff:
@dulciejackson I found a few more files to remove - can you review?

jamesgriff (approved):

dulciejackson:
👍 @jamesgriff the extras look good to me!

PR #257, opened 2020-10-14
GPG-418 Update Email Unit Tests

dulciejackson:
The code for this was already done (woop woop) so I just added a few unit tests to make sure it all works ok 😄

### DIFF
--- a/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Account/ChangeEmailTests.cs
+++ b/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Account/ChangeEmailTests.cs
@@ -1,0 +1,100 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using GenderPayGap.Database;
+using GenderPayGap.WebUI.Controllers.Account;
+using GenderPayGap.WebUI.Models.Account;
+using GenderPayGap.WebUI.Services;
+using GenderPayGap.WebUI.Tests.Builders;
+using Microsoft.Extensions.Primitives;
+using NUnit.Framework;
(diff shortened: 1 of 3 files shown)


GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Account/ChangeEmailTests.cs
```diff
@@ -0,0 +1,232 @@
...
+    [SetCulture("en-GB")]
+    public class ChangeEmailTests
+    {
+
+        [Test]
+        [Description("POST: Providing a valid new email address results in a verification email being sent")]
+        public void POST_Providing_A_Valid_New_Email_Address_Results_In_A_Verification_Email_Being_Sent()
+        {
+            // Arrange
+            User user = new UserBuilder().WithEmailAddress("old@test.com").Build();
```
jamesgriff, line 28: This is really minor, but I tend to use `@example.com` for fake email addresses:
* just in case we accidentally send the email address a real email
* because http://example.com is an official domain for use in documentation, so it there won't be a real person who might end up receiving our emails

jamesgriff (approved):
An optional change, but otherwise looks great - approved :-)

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/Controllers/Account/ChangeEmailTests.cs
dulciejackson: Ah that's interesting, I didn't know that! Will swap them over 😄

jamesgriff (approved):

PR #259, opened 2020-10-19
GPG-387 (GPG-524 and 526) Update "Manage Organisations" Pages

dulciejackson:
The first chunk of functionality for the ManageOrganisations journey!

This:
- Creates a new ManageOrganisationsController
- Creates new ManageOrganisationsViewModel and ManageOrganisationViewModel
- Implements functionality for ManageOrganisations and ManageOrganisation GET endpoints
- Creates replacement views with new design system styling
- Adds feature flag for Launchpad.cshtml redirect on "submit gender pay gap data" link
- Adds OrganisationReportsSection.cshtml to replace DisplayReportSection with new design system styling
- Adds UsersRegisteredToReportForOrganisation.cshtml to separate out the ManageOrganisation.cshtml file a bit (in keeping with the above reports table)


GenderPayGap.WebUI/Controllers/Guidance/EightWaysGuidanceController.cs
```diff
@@ -74,7 +74,7 @@ public static PaginationPages GetPagesInThisGuidance(IUrlHelper helper, HttpRequ
                 },
                 new PaginationPage
                 {
-                    Title = "Do people get �stuck� at certain levels within your organisation?",
+                    Title = "Do people get ‘stuck’ at certain levels within your organisation?",
```
dulciejackson, line 77: IntelliJ automatically did this - I think it updated the characters used for the single quotes (even though it doesn't look like it)!

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisations.cshtml
```diff
@@ -0,0 +1,82 @@
...
+                    } else if (userOrganisation.PINSentDate == null && userOrganisation.PINConfirmedDate == null)
+                    {
+                        statusText = "Awaiting registration approval";
+                        linkAction = null;
+                    }
+                    else
+                    {
+                        statusText = "Registration complete";
+                        linkAction = Url.Action("ManageOrganisationGet", "ManageOrganisations", new {id = encryptedOrganisationId});
+                    }
```
dulciejackson, line 66: I wasn't sure if this chunk of C# was better than the old version (in `Views/Organisation/ManageOrganisations.cshtml`) but it separated it out a bit from lots of processing alongside lots of html?

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
```diff
@@ -0,0 +1,124 @@
...
+            <th scope="col" class="govuk-table__header">Year</th>
+            <th scope="col" class="govuk-table__header">Reporting requirement</th>
+            <th scope="col" class="govuk-table__header">Report status</th>
+        </tr>
+    </thead>
+    <tbody class="govuk-table__body">
+        @{
+            List<int> reportingYears = ReportingYearsHelper.GetReportingYears();
+
+            for (int i = 0; i < reportingYears.Count; i++)
```
dulciejackson, line 20: I feel like this file is horrible - it feels like a real code smell to me that there's _so much_ C# alongside the HTML here. I tried to refactor some of the complexity out, but I think a lot of it is pretty much essential so this uses the right one of the many complicated variable urls/bits of text in each place!

GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            this.dataRepository = dataRepository;
+        }
+
+        [HttpGet]
+        public IActionResult ManageOrganisationsGet()
+        {
+            User user = ControllerHelper.GetGpgUserFromAspNetUser(User, dataRepository);
+            
+            // Show the privacy policy to non-admin users (if they're not being impersonated) if they haven't read it yet
```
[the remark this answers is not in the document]
dulciejackson: As far as I can tell, this isn't used elsewhere at the moment - although it seems slightly odd to me that we'd only show the privacy policy here and not elsewhere? 

@jamesgriff do you think this would be useful?

GenderPayGap.WebUI/Models/Organisation/OrganisationReportsViewModel.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        private Dictionary<int, OrganisationScopeForYear> GetOrganisationScopes(Database.Organisation organisation, List<int> yearsWithDraftReturns)
+        {
+            // Get list of all available reporting years
+            List<int> reportingYears = ReportingYearsHelper.GetReportingYears();
+            
+            // Create dictionary to store reporting years as keys against OrganisationScopeForYear values
+            Dictionary<int, OrganisationScopeForYear> scopesByYear = new Dictionary<int, OrganisationScopeForYear>();
+            
+            // Iterate over each available reporting year
```
Binney, line 36: There's a few comments here I find overkill - no particular downside to having them but I definitely don't think they're needed and there's a case for deleting them just to keep the lines in the file down

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
```diff
@@ -0,0 +1,100 @@
...
+    <div class="govuk-grid-column-full">
+        <span class="govuk-caption-xl">Reporting as @Model.Organisation.OrganisationName</span>
+        <h1 class="govuk-heading-xl">Manage your organisation's reporting</h1>
+        
+        <dl>
+            <dt class="govuk-body govuk-!-font-weight-bold">
+                Address
+            </dt>
+            <div>
+                <dt class="govuk-caption-m govuk-!-font-size-16">
```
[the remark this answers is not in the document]
dulciejackson: Yup, I copied and pasted too fast!

GenderPayGap.WebUI/Models/Organisation/OrganisationReportsViewModel.cs
```diff
@@ -0,0 +1,156 @@
...
+            {
+                return "Edit report";
+            }
+
+            return "Edit draft report";
+        }
+
+        public DateTime GetSnapshotDate()
+        {
+            return this.snapshotDateForYear;
```
Binney, line 152: Does this need to be here? can you expose a getter on `snapshotDateForYear` directly?

GenderPayGap.WebUI/Views/Viewing/Launchpad/Index.cshtml
```diff
@@ -38,11 +38,17 @@
     Report your gender pay gap data
 </div>
 
+@{
+    var reportingUrl = FeatureFlagHelper.IsFeatureEnabled(FeatureFlag.NewManageOrganisationsJourney)
```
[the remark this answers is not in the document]
dulciejackson: I think we'd normally look through uses of the feature flag and refactor to point to just the "new" version, and then remove the feature flag entirely :)

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisations.cshtml
```diff
@@ -0,0 +1,104 @@
...
+            <thead class="govuk-table__head">
+                <tr class="govuk-table__row">
+                    <th scope="col" class="govuk-table__header">Organisation name</th>
+                    <th scope="col" class="govuk-table__header">Registration status</th>
+                </tr>
+            </thead>
+            <tbody class="govuk-table__body">
+                @foreach (UserOrganisation userOrganisation in Model.UserOrganisations)
+                {
+                    string statusText;
```
Binney, line 50: I'm deeply confused about how Github is showing me this diff but these appear to still be present and not be needed any more

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisations.cshtml
```diff
@@ -0,0 +1,104 @@
...
+                        {
+                            <td class="govuk-table__cell">
+                                <a class="govuk-link" href="@Url.Action("ManageOrganisationGet", "ManageOrganisations", new {id = encryptedOrganisationId})">
+                                    @(userOrganisation.Organisation.OrganisationName.ToUpper())
+                                </a>
+                            </td>
+                            <td class="govuk-table__cell">
+                                Registration complete
+                            </td>    
+                        }n
```
Binney, line 98: `n`?

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
```diff
@@ -0,0 +1,83 @@
...
+    <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+            <th scope="col" class="govuk-table__header">Year</th>
+            <th scope="col" class="govuk-table__header">Reporting requirement</th>
+            <th scope="col" class="govuk-table__header">Report status</th>
+        </tr>
+    </thead>
+    <tbody class="govuk-table__body">
+        @{
+            foreach (KeyValuePair<int, OrganisationScopeForYear> scopeForYear in Model.OrganisationScopesByYear)
```
Binney, line 16: Since this is just a `foreach`, could we just do it as a list of `OrganisationScopeForYear`? (would also make the generation of Model.OrganisationScopesByYear a little easier)

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
```diff
@@ -0,0 +1,83 @@
...
+            {
+                int reportingYear = scopeForYear.Key;
+                OrganisationScopeForYear organisationScopeForYear = scopeForYear.Value;
+
+                string formattedYearText = $"{reportingYear}/{(reportingYear + 1).ToTwoDigitYear()}";
```
Binney, line 21: This should also go in the viewmodel

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
```diff
@@ -0,0 +1,83 @@
...
+                    </td>
+                    
+                    <td class="govuk-table__cell">
+                        <div>
+                            @organisationScopeForYear.GetReportStatusText()
+                        </div>
+                        <div>
+                            @if (FeatureFlagHelper.IsFeatureEnabled(FeatureFlag.NewReportingJourney))
+                            {
+                                <a class="govuk-button govuk-!-margin-top-4 govuk-!-margin-bottom-0" href="@Url.Action("ReportOverview", "ReportOverview",
```
Binney, line 64: Github is complaining about these, which I think is inadvertently because it expects each HTML attribute to be on only one line - thoughts? i.e. a line break before the `href`

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
```diff
@@ -0,0 +1,83 @@
...
+        @{
+            foreach (KeyValuePair<int, OrganisationScopeForYear> scopeForYear in Model.OrganisationScopesByYear)
+            {
+                int reportingYear = scopeForYear.Key;
+                OrganisationScopeForYear organisationScopeForYear = scopeForYear.Value;
+
+                string formattedYearText = $"{reportingYear}/{(reportingYear + 1).ToTwoDigitYear()}";
+                
+                DateTime snapshotDate = organisationScopeForYear.GetSnapshotDate();
+                DateTime deadline = snapshotDate.AddYears(1).AddDays(-1);
```
Binney, line 24: Could this go in the viewmodel too? (Instead of having a local var at all I mean - making `GetByReportingDeadlineText` take no arguments)

GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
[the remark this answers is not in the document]
Binney: I like it! I agree with Dulcie, it does feel odd we only show the privacy policy here - but I think that's a problem with the business logic not with the code, so pulling it out feels right to me

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
```diff
@@ -0,0 +1,100 @@
...
+        @{
+            await Html.RenderPartialAsync("../Components/CovidMessage");
+        }
+    </div>
+}
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+        <span class="govuk-caption-xl">Reporting as @Model.Organisation.OrganisationName</span>
+        <h1 class="govuk-heading-xl">Manage your organisation's reporting</h1>
```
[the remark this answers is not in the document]
dulciejackson: Ok, I just went with what the design system suggested for this sort of case - is there a reason the admin site doesn't use that styling?

GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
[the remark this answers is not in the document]
dulciejackson: Clever!
Is it semantically "good practice" for an exception to redirect to an actual page rather than an error page of some kind? Not questioning it, just curious!

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
```diff
@@ -0,0 +1,100 @@
...
+        
+        <dl>
+            <dt class="govuk-body govuk-!-font-weight-bold">
+                Address
```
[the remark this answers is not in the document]
dulciejackson: This has been bundled into a whole grouping of other design work (e.g. Manage Organisation/Account tabs) that Claire's going to get to at some point and then feed into more specific dev tickets. For now it's been agreed with Jacob to build this exactly as it is right now to stop this being blocked by the need for more design!

GenderPayGap.WebUI/Views/Viewing/Launchpad/Index.cshtml
```diff
@@ -38,11 +38,17 @@
     Report your gender pay gap data
 </div>
 
+@{
+    var reportingUrl = FeatureFlagHelper.IsFeatureEnabled(FeatureFlag.NewManageOrganisationsJourney) 
+        ? Url.Action("ManageOrganisationsGet", "ManageOrganisations") 
+        : Url.Action(nameof(OrganisationController.ManageOrganisations), "Organisation");
```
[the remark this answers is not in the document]
dulciejackson: I've added a feature flag check in each of the old and new manage organisations endpoints to redirect to the other one if the feature flag is enabled/disabled respectively - so a user can't manually input the new journey URL when it's not meant to be live and vice versa. [GPG-517](https://technologyprogramme.atlassian.net/browse/GPG-527) is in the pipeline to swap all other links over to pointing at the right one (although in practice I don't think it'll make a difference if we're checking the feature flag in both directions)

GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
```diff
@@ -0,0 +1,90 @@
...
+        private readonly IDataRepository dataRepository;
+
+        public ManageOrganisationsController(IDataRepository dataRepository)
+        {
+            this.dataRepository = dataRepository;
+        }
+
+        [HttpGet]
+        public IActionResult ManageOrganisationsGet()
+        {
```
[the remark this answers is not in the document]
dulciejackson: Do we want to give a 404, or just automatically redirect to the "right" version of the page? I imagine it'd be confusing to e.g. follow a link to Manage Organisations, it have swapped to the new version on live, and you get a 404 rather than the controller leading to the right place!

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
dulciejackson: How I didn't think of this, I have no idea 😅

GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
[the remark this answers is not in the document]
Binney: Well I sort of like "exception as fall-through behaviour" - notwithstanding it's sort of the only way we'd do this particular thing - an exception shouldn't just be "something has gone wrong" but "something has happened that's an exception to the usual rules of what is going on" - no reason that behaviour should be pathological. This is quite similar to "redirect to a signin page if you're not logged in", certainly it's a pattern we use across BBC Monitoring (where the permissions system is multi-layered)

GenderPayGap.WebUI/Models/Organisation/OrganisationReportsViewModel.cs
```diff
@@ -17,29 +17,25 @@ public class OrganisationReportsViewModel
...
         {
-            // Get list of all available reporting years
             List<int> reportingYears = ReportingYearsHelper.GetReportingYears();
             
-            // Create dictionary to store reporting years as keys against OrganisationScopeForYear values
-            Dictionary<int, OrganisationScopeForYear> scopesByYear = new Dictionary<int, OrganisationScopeForYear>();
+            List<OrganisationScopeForYear> scopesByYear = new List<OrganisationScopeForYear>();
             
-            // Iterate over each available reporting year
             foreach (int reportingYear in reportingYears)
```
Binney, line 34: Fun fact: you can do this in a super functional way using Linq (e.g. `reportingYears.Select(organisation.GetScopeForYear)`, filter on not-nulls, then Select to create the OrganisationScopeForYear) but this is probably clearer if we're not doing things the functional way elsewhere

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
```diff
@@ -10,8 +12,24 @@
 
 @section BeforeMain {
     @{
-        await Html.RenderPartialAsync("../Components/Navigation/AccountTabs");
+        var crumbs = new List<CrumbViewModel>
```
Binney, line 15: crumbs

GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
```diff
@@ -73,6 +65,7 @@ public IActionResult ManageOrganisationGet(string id)
             }
```
Binney, line 65: (I've lost track of what's already on this branch/review/PR but) I think there's a helper for this (the "404 if decryption fails")

Binney (approved):
Niceeeeeee

GenderPayGap.WebUI/Models/Organisation/OrganisationReportsViewModel.cs
dulciejackson: Ah yeah, I love these functional Linq extras!
I'm really keen on functional stuff, but aware that that's not classic C# style and might be a bit unfamiliar to people with a C# background picking up maintenance in future. Also aware that other places use `foreach`, so trying to keep it consistent! :)

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
dulciejackson: 🍞

GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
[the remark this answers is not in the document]
jamesgriff: For these pages, your solution works nicely.
For other pages, I think a 404 would be appropriate (e.g. if it's a new feature where the public shouldn't know that the feature exists yet).
It can sometimes get a bit messy trying to work out what the best equivalent page is (e.g. if the structure of the journey has changed) so sometimes a 404 is cleanest - but it depends on the particular case.
I think this redirect-one-way-or-the-other works nicely here.

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
jamesgriff: I like the order:
```cshtml
<a href="..."
   class="govuk-link">
    Link text
</a>
```
It helps me remember that there should always be a `class="govuk-link"` on each `<a>` tag (except in this case where it's a button)

jamesgriff: Actually, styling links as buttons is frowned upon ☹️ for accessibility reasons (happy to have a call to discuss more details)
If these are going to be buttons, then it might be better if each one is inside a form:
```cshtml
<form method="get"
      action="@(Url.Action(...))">
    <button .../>
</form>
```
I know that's a bit of a bore, but it does genuinely help people using speech-to-command software

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisations.cshtml
```diff
@@ -0,0 +1,104 @@
...
+    
+    <div class="govuk-!-margin-top-3">
+        @{
+            await Html.RenderPartialAsync("../Components/CovidMessage");
+        }
+    </div>
+}
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
```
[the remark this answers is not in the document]
dulciejackson: Yeah, this was me mis-typing on impulse! The original (and designs) use full-width, so I've reverted back to that and it should look better now!

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisations.cshtml
```diff
@@ -0,0 +1,104 @@
...
+                            </td>
+                            <td class="govuk-table__cell">
+                                Registration complete
+                            </td>    
+                        }
+                    </tr>
+                }
+            </tbody>
+        </table>
+    </div>
```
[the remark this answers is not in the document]
dulciejackson: I'm not sure what to do about the conditional bit of text here; the "If you created an account by mistake or no longer responsible for reporting gender pay gap data you can close your account..." - I'm not quite sure how to determine when to show it?

jamesgriff (commented):
I've addressed some of my comments in https://github.com/cabinetoffice/gender-pay-gap/pull/268

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisations.cshtml
```diff
@@ -84,5 +83,21 @@
...
+        
+        @(Html.GovUkButton(new ButtonViewModel
+        {
+            Text = "Add organisation",
+            Classes = "govuk-!-margin-bottom-0",
+            Href = Url.Action("ChooseSector", "AddOrganisationChooseSector")
+        }))
+        
+        @{
+            // TODO: do something sensible to show the close account text
```
Binney, line 95: Do we have a way of tracking TODOs anywhere? All else being equal this should at least tag the ticket number i.e. GPG-387

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
dulciejackson: Could the `aria-role` come in handy here? That way we could keep the links, style them as buttons but a screen-reader would present them as buttons?

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisations.cshtml
dulciejackson: Ah this is a TODO until I figure out how to do it/if it's necessary in this PR! Just there to remind myself to come back to before merging, but I doubt it warrants a whole new ticket :)

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
jamesgriff: Ooh yes, I think that would work - great thought 😄

GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisations.cshtml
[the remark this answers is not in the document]
jamesgriff: I think this will be
```cshtml
@if (Model.UserOrganisations.Count == 0)
{
    ..."If you created your account"...
}
```

GenderPayGap.WebUI/Models/Organisation/OrganisationReportsViewModel.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                return "Edit report";
+            }
+
+            return "Edit draft report";
+        }
+
+        public string GetFormattedYearText()
+        {
+            return $"{ReportingYear}/{(ReportingYear + 1).ToTwoDigitYear()}";
```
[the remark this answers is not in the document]
dulciejackson: I've added a ticket to refactor this into a helper method and update all references!

jamesgriff:
@dulciejackson 
I've added a few commits - can you review?
Other than that, I'm happy for it to be merged 😄

jamesgriff (approved):

PR #261, opened 2020-10-20
GPG-529 Remove confusing email behaviour

jamesgriff:
In test environments, "registration approved/rejected" emails are sent to the admin instead of the normal user.
This behaviour is very confusing and not needed (since we have different Notify API keys for test/live)


Binney:
Good shout, I agree this is confusing.

Previously we had two layers of protection against sending real users emails:
* this, and
* the different Notify key (what exactly does this do?)
All else being equal I'd like to have two layers for things like this. is there anything else we can add in e.g. in the emailSendingService itself? Or can you convince me the Notify key is enough on its own?

Binney (approved):

dulciejackson (approved):

PR #262, opened 2020-10-20
GPG-387 Extract "Redirect to Privacy Policy" into one-line method call

jamesgriff:


GenderPayGap.WebUI/Helpers/ControllerHelper.cs
```diff
@@ -115,5 +116,18 @@ public static void ThrowIfAdminIsImpersonatingUser(ClaimsPrincipal aspDotNetUser
...
 
+        public static void RedirectIfUserNeedsToReadPrivacyPolicy(ClaimsPrincipal aspDotNetUser, User gpgUser, IUrlHelper url)
+        {
+            // Show the privacy policy to non-admin users (if they're not being impersonated) if they haven't read it yet
+            if (!LoginHelper.IsUserBeingImpersonated(aspDotNetUser) && !gpgUser.IsAdministrator())
+            {
+                DateTime? hasReadPrivacy = gpgUser.AcceptedPrivacyStatement;
+                if (hasReadPrivacy == null || hasReadPrivacy.Value < Global.PrivacyChangedDate)
+                {
+                    throw new RedirectToPrivacyPolicyException(url);
```
Binney, line 127: (Not particularly relevant to this change but) should we be passing through the URL of the page you originally wanted to go to? I can see in `PrivacyPolicyController` you always redirect back to `ManageOrganisations` but I assumed on reading this that was what `url` was for so it's a little surprising

Binney (approved):
Niceeee

GenderPayGap.WebUI/Helpers/ControllerHelper.cs
jamesgriff: That would be nice
If we start redirecting to the Privacy Policy from other pages, then I agree we should also redirect back to the page that they came from
Perhaps this should be a separate ticket?

PR #264, opened 2020-10-22
GPG-534 In test environments, send all emails to a test Gmail account

jamesgriff:


GenderPayGap.WebUI/Services/EmailSendingService.cs
```diff
@@ -393,24 +349,23 @@ public async void SendAccountRetirementNotificationEmail(string emailAddress, st
                 Personalisation = personalisation
             };
 
-            await AddEmailToQueue(notifyEmail);
+            AddEmailToQueue(notifyEmail);
         }
 
-        private async Task<bool> AddEmailToQueue(NotifyEmail notifyEmail)
+        private void AddEmailToQueue(NotifyEmail notifyEmail)
```
Binney, line 355: Great spot on the async :D

Binney (approved):
Epic, that was quick!

What changes do we need to make to the Notify templates for this? Do we have a way of keeping track of those templates so we remember to apply them directly after release? (e.g. a release ticket)

jamesgriff:
@Binney in the end we didn't need any changes to the templates 😄 we're adding the email address to the existing `Environment` parameter - at some point it might be nice to give it a clearer name, but we'd need to re-create all the templates, which is a bit tedious!

PR #266, opened 2020-10-22
GPG-467 Fixes unmarked headings on add org page

dulciejackson:
### Context
Users who cannot perceive visually styling rely on correct semantic mark-up in order to identify information and relationships within the page. Currently the company name headings are marked up with a `<span>` rather than headings, which would make it hard for users of screen-readers to filter and efficiently navigate the page.

### Changes
Replaced `<span>` with styled `<h3>` tags

### DIFF
--- a/GenderPayGap.WebUI/Views/AddOrganisationSearch/Search.cshtml
+++ b/GenderPayGap.WebUI/Views/AddOrganisationSearch/Search.cshtml
@@ -196,9 +196,9 @@
                             : $"search-result-coho-{organisation.CompanyNumber}";
                         <li id="@(resultId)">
                             <div>
-                                <span class="govuk-!-font-weight-bold">
+                                <h3 class="govuk-heading-s govuk-!-margin-bottom-0">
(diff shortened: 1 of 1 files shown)


GenderPayGap.WebUI/Views/AddOrganisationSearch/Search.cshtml
```diff
@@ -196,9 +196,9 @@
                             : $"search-result-coho-{organisation.CompanyNumber}";
                         <li id="@(resultId)">
                             <div>
-                                <span class="govuk-!-font-weight-bold">
+                                <h3 class="govuk-heading-s govuk-!-margin-bottom-0">
                                     @(organisation.OrganisationName)
-                                </span>
+                                </h3>
                                 <br/>
```
jamesgriff, line 202: There's now a big gap under each organisation name
If we remove the `<br/>` then it looks good

Binney (approved):

jamesgriff (approved):

PR #268, opened 2020-10-22
GPG-387 Suggestions for refactoring the Manage Organisation sub-pages / view-models

jamesgriff:
Hi @dulciejackson,

Here's some suggestions for the Manage Organisation sub-pages and view-models.
Feel free to use them / ignore them as you like (apart from CanChangeScope where the logic is different).

Some reasons for the changes:
* The logic for CanChangeScope needed a change, which was going to result in a few changes that made the code a bit ugly
* I find the code can get confusing when view-models are created within views (rather than in controllers or other views)
* I find methods easier to follow than properties that are set in a constructor (I tend to do a "Find Usages" on properties to see where they're set - if they have setters, then there's a risk the code will set them in surprising places)
* The view-models are all used only on the 1 page and had quite a lot of overlapping fields - so re-using a view-model seems sensible
* There was quite a lot of data being passed around (e.g. the OrganisationScopeForYear constructor took 5 parameters (and needed a 6th)) - but it could be simpler if we pass a bigger object (Organisation) and get the data later when we need it

Let me know what you think.
😄


Binney (approved):

GenderPayGap.WebUI/Models/Organisation/ManageOrganisationViewModel.cs
```diff
@@ -1,21 +1,169 @@
...
+            this.organisation = organisation;
+            ReportingYear = reportingYear;
+            this.hasDraftReturnForYear = hasDraftReturnForYear;
+        }
+
+
+        public object GetFormattedYearText()
+        {
+            return $"{ReportingYear}-{(ReportingYear + 1)%100}";
+        }
```
dulciejackson, line 86: Would this be better in a helper somewhere, since we seem to use this all the time?

GenderPayGap.WebUI/Models/Organisation/ManageOrganisationViewModel.cs
```diff
@@ -1,21 +1,169 @@
...
+
+        public bool CanChangeScope()
+        {
+            int currentReportingYear = organisation.SectorType.GetAccountingStartDate().Year;
+            int earliestAllowedReportingYear = currentReportingYear - (Global.EditableScopeCount - 1);
+            bool isCurrentYearEditable = ReportingYear >= earliestAllowedReportingYear;
+            return isCurrentYearEditable;
```
dulciejackson, line 93: Literally just personal preference, but I'd find this more readable if it were just
`return ReportingYear >= earliestAllowedReportingYear;`

Thoughts? Happy with either if you find this clearer 😄

dulciejackson (commented):
Looks great - agree using one Model is much cleaner than creating multiple in the Views! Just a couple of entirely optional points :)

GenderPayGap.WebUI/Models/Organisation/ManageOrganisationViewModel.cs
jamesgriff: Yes, let's put it in `ReportingYearHelper`

GenderPayGap.WebUI/Models/Organisation/ManageOrganisationViewModel.cs
jamesgriff: I'm happy either way 😄

GenderPayGap.WebUI/Models/Organisation/ManageOrganisationViewModel.cs
jamesgriff: Ooh, actually, there's already a method - `ReportingPeriodHelper.FormatReportingPeriod`

dulciejackson (approved):
Looks great! Thanks for picking up this stuff 😄

PR #276, opened 2020-11-03
GPG-469 Removes ? icons and links and replaces with details sections

dulciejackson:
[GPG-469 JIRA Ticket](https://technologyprogramme.atlassian.net/browse/GPG-469)
[What on earth I'm talking about](https://gender-pay-gap.service.gov.uk/compare-employers) (you'll need to add an org to the compare for this to work)

### Context
DAC mentioned an accessibility issue with the compare table headers, where the undescriptive aria-label was overriding the useful visibly hidden text so it was unclear what happened with the ? icons. Originally these icons opened a new window to show extra information about mean and median or quartiles, which could be confusing as it wasn't announced for a screen-reader.

### Changes
A screen-reader will read out all nested visible text and aria-labels in one concatenated "sentence" for a table header, which it reads both on a first scan of the headings *and* for each table entry. This means that it's a huge pain to make it say something sensible to include both the header text and the help icon/link, and then not read out a really long thing for every item in the table column. 

Because of this, I've removed the ? icons entirely and swapped to using the GOV.UK Design System Details components at the top of the page with the "How can employers close the gap" link.


GenderPayGap.WebUI/Views/Compare/CompareEmployers.cshtml
```diff
@@ -20,10 +20,35 @@
     </div>
 
     <div class="grid-row">
-        <div class="column-two-thirds">
+        <div class="column-two-thirds govuk-!-margin-bottom-2">
             <a class="link-button" rel="track" data-track-category="Related Link Clicked" href="/actions-to-close-the-gap">How can employers close the gap</a>
-            <br/>
-            <br/>
+        </div>
+        <div class="column-full">
```
jamesgriff, line 26: I think `column-two-thirds` might look nicer here (I find the shorter lines of text easier to read)

jamesgriff (commented):
One suggestion (happy for you to ignore if you disagree)
And I added a commit with a few files we can remove

jamesgriff (approved):

PR #279, opened 2020-11-06
GPG-390 Action Hub Accessibility

dulciejackson:
[GPG-390/549 JIRA Ticket](https://technologyprogramme.atlassian.net/browse/GPG-549)

**Part 1/2**
This PR updates the ActionHubOld views to the new design system - another one will update the ActionHubNew views.

**Update: It's now been agreed to postpone updating the ActionHubNew work until it's needed (if ever)**

### Context
The old design system components are inaccessible in some ways, so we want to update to the new design system to get rid of as many accessibility problems as possible!


GenderPayGap.WebUI/wwwroot/styles/components/_guidance.scss
```diff
@@ -60,3 +60,22 @@
         }
     }
 }
+
+.print-link {
+    background: url(../../img/icon-print.png) no-repeat 0 50%;
+    padding: 0.5em 0 0.5em 28px
```
Binney, line 66: Having a mix of `em` and `px` feels wrong to me. Also, the `28px` value for the left-padding should depend on the size of the icon, and therefore the font size. can you convert it to em?

GenderPayGap.WebUI/Views/ActionHubOld/MixedResult.cshtml
```diff
@@ -0,0 +1,142 @@
...
+    {
+        Title = "Previous",
+        Text = "Promising actions",
+        Url = Url.Action("Promising", "ActionHub")
+    };
+
+    await Html.RenderPartialAsync("/Views/Components/PaginationLinks/PaginationLinks.cshtml", new PaginationLinksViewModel
+    {
+        LeftLink = leftLink,
+        RightLink = null
```
Binney, line 134: it really bugs me how this drops you off a cliff and doesn't give you any logical onward journeys... again not one for this PR though! :D

Binney (changes requested):
Mostly CSS tweaks :)

GenderPayGap.WebUI/Views/ActionHubOld/MixedResult.cshtml
dulciejackson: Agreed! It's a bit weird, but I think the chunkier redesign work will happen at some point soon-ish and that's when stuff like this will be tackled! 😄

GenderPayGap.WebUI/Views/ActionHubOld/Promising.cshtml
```diff
@@ -0,0 +1,189 @@
...
+            </strong>
+        </p>
+    <p class="govuk-body">
+        The gender pay gap widens dramatically after women have children but this could be reduced if men and women were able to share childcare more equally.
+        Shared Parental Leave and Pay enables working parents to share up to 50 weeks of leave and up to 37 weeks of pay in their child’s first year.
+    </p>
+    <ul class="govuk-list govuk-list--bullet">
+        <li>Offer enhanced Shared Parental Pay at the same level as enhanced maternity pay</li>
+        <li>
+            Encourage take up of Shared Parental Leave (see our <a target="_blank" rel="noopener" href="https://sharedparentalleave.campaign.gov.uk/"><span class="visually-hidden">Shared Parental Leave </span>guidance<span class="visually-hidden">(opens in a new window)</span></a>). For example:
```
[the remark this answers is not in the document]
dulciejackson: Ah yeah sorry, mistyped that one!

GenderPayGap.WebUI/Views/ActionHubOld/Promising.cshtml
```diff
@@ -0,0 +1,189 @@
...
+            <ul class="govuk-list govuk-list--bullet">
```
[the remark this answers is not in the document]
dulciejackson: Agreed - design-wise I was a bit unsure what to do here because the old version used dots for top-level list items, and open circles for the second-level items. I could recreate that with some custom CSS, but given that the design system uses the same list bullet for all levels as default I thought saving the code and keeping it consistent was good enough rationale to not bother?

GenderPayGap.WebUI/wwwroot/styles/components/_number-icons.scss
```diff
@@ -0,0 +1,37 @@
...
+
+.action-step .icon {
+  position: absolute;
+}
+
+.action-step h3,
+.action-step p,
+.action-step ul {
+  padding-left: 30px;
+}
```
jamesgriff, line 14: A couple of thoughts:
* Since we're using Sass, we can nest this to make it really clear that the `h3`/`p`/`ul` are within the `.action-step`, like this:
   ```scss
    .action-step {
        position: relative;
        padding-bottom: 10px;

        h3, p, ul {
            padding-left: 30px;
        }
    }
    ```
    In fact, it would be great if **all the CSS** in this file was within `.action-step` so that we don't accidentally re-use it somewhere else (e.g. it would be easy to use the class `icon` somewhere without remembering that it's defined here)

* But, even better, instead of having to list all the elements (`h3`, `p`, `ul`) that we happen to have inside the `.action-step`, it would be nice it we could put anything inside the `.action-step` - e.g.
    ```scss
    .action-step {
        position: relative;
        padding-bottom: 10px;
        padding-left: -30px; // new

        // can remove this section
        //h3, p, ul {
        //   padding-left: 30px;
        //}

        .icon {
            position: absolute;
            margin-left: -30px; // new
        }
    }
    ```

GenderPayGap.WebUI/wwwroot/styles/components/_guidance.scss
```diff
@@ -60,3 +60,28 @@
         }
     }
 }
+
+.print-link {
+    background: url(../../img/icon-print.png) no-repeat 0 50%;
```
jamesgriff, line 65: If there's any images we're using on new-style pages, it would be great to put them in the `wwwroot/assets/images` folder (so it's easier to see which are old and which are new)

GenderPayGap.WebUI/Views/ActionHubOld/Promising.cshtml
```diff
@@ -0,0 +1,189 @@
...
+        <div class="govuk-accordion__section-header">
+            <h2 class="govuk-accordion__section-heading">
+                <span class="govuk-accordion__section-button" id="accordion-default-heading-1">
+                    References
+                </span>
+            </h2>
+        </div>
+        <div id="accordion-default-content-1" class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-1">
+            @{
+                var referenceIndex = 1;
```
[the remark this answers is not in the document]
jamesgriff: I think that, semantically, this whole section should be within an `<ol>`.
Each `<ol>` can contain a `<ul>` if we want sub-lists
A benefit would be that you get the numbering for free

GenderPayGap.WebUI/Views/ActionHubOld/Index.cshtml
```diff
@@ -0,0 +1,90 @@
+﻿@{
```
jamesgriff, line 1: I think this file should have "layout" in it's name, rather than "index".
Index implies it's the homepage of this folder.

In the `EightWaysGuidance` folder, we've gone for a structure:
* Pages
  * SomePage.cshtml
  * AnotherPage.cshtml
* EightWaysGuidanceLayout.cshtml

What do you think about using a similar folder/file structure here?

GenderPayGap.WebUI/Views/ActionHubOld/Promising.cshtml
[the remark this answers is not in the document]
dulciejackson: Do you think this is ok? (The numbers aren't bold anymore)
image

dulciejackson:
I've updated the SCSS for the action steps to style an `<ol>` to match the references and help visually impaired users. Originally the action step classes were applied only visually by adding e.g. `icon-step-1` class to something, whereas now the numbered list will be interpreted correctly by a screen-reader and the references will follow the same format. Hopefully this is ok?

GenderPayGap.WebUI/wwwroot/styles/components/_number-icons.scss
```diff
@@ -0,0 +1,28 @@
...
+    font-size: small;
+    text-align: center;
+    vertical-align: middle;
+    width: 23px;
+    height: 23px;
+    left: -30px;
+    color: white;
+    background: black;
+    border-radius: 50%;
+  }
```
jamesgriff, line 27: This is all _very cool_ 😎 but I wonder if it would be easier to read & more in-keeping with Design System styling if we use a more normal number styling.

e.g. just making the numbers bold:
```scss
ol > li::marker {
    font-weight: bold;
}
```

If you'd prefer to keep this styling, there's a small bug with how it looks on mobile (the numbers fall off the left of the screen)

GenderPayGap.WebUI/Views/ActionHubOld/Promising.cshtml
```diff
@@ -0,0 +1,187 @@
...
+            <h2 class="govuk-accordion__section-heading">
+                <span class="govuk-accordion__section-button" id="accordion-default-heading-1">
+                    References
+                </span>
+            </h2>
+        </div>
+        <div id="accordion-default-content-1" class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-1">
+            <div>
+                <ol class="govuk-list govuk-list--number">
+                    <li>
```
jamesgriff, line 116: This new `<ol>` / `<li>` structure for the References section looks much cleaner 😄

GenderPayGap.WebUI/Views/ActionHubOld/MixedResult.cshtml
```diff
@@ -0,0 +1,142 @@
...
+                    References
+                </span>
+            </h2>
+        </div>
+        <div id="accordion-default-content-1" class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-1">
+            @{
+                var referenceIndex = 1;
+            }
+            <div>
+                <h4 class="govuk-heading-s">@(referenceIndex++). Unconscious bias training</h4>
```
jamesgriff, line 70: Could you use the new `<ol>` / `<li>` structure for the References section on this page?

GenderPayGap.WebUI/Views/ActionHubOld/Effective.cshtml
```diff
@@ -0,0 +1,200 @@
...
+                    References
+                </span>
+            </h2>
+        </div>
+        <div id="accordion-default-content-1" class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-1">
+            @{
+                var referenceIndex = 1;
+            }
+            <div>
+                <h4 class="govuk-heading-s">@(referenceIndex++). Include multiple women in shortlists for recruitment and promotions</h4>
```
jamesgriff, line 119: Could you use the new `<ol>` / `<li>` structure for the References section on this page?

GenderPayGap.WebUI/Views/ActionHubOld/Promising.cshtml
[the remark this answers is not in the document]
jamesgriff: I think it's ok either way.
If you do want to make them bold, you can add some CSS:
```scss
.my-ol-class > li::marker {
    font-weight: bold;
}
```

GenderPayGap.WebUI/wwwroot/styles/components/_number-icons.scss
dulciejackson: I thought about this - but a lot of the guidance has been created with various versions of these icons (here they're numbers, on the new version there were replacement image icons) so I wasn't sure if it was likely that we'd want to keep this styling in the case of an eventual redesign? Happy to go back to just bold numbers, but maybe for now it's a question for Claire r.e. style since it won't make much difference technically?

jamesgriff: Yeh sure - happy to leave them as numbers in circles.
We can present the "just bold numbers" as an alternative to Claire when she looks at this in future.
It's good that we know how to do this either way.

There's a remaining small bug on narrow screens (the numbers fall off the left of the screen) - I expect some margin/padding somewhere could fix this though

dulciejackson: (I've fixed the numbers falling off the page on mobile for now 😄)

jamesgriff (approved):

GenderPayGap.WebUI/wwwroot/styles/components/_number-icons.scss
```diff
@@ -0,0 +1,37 @@
...
+    content: counter(action-step-counter);
+    position: absolute;
+    line-height: 23px;
+    font-size: small;
+    text-align: center;
+    vertical-align: middle;
+    width: 23px;
+    height: 23px;
+    color: white;
+    background: black;
```
Binney, line 30: If you wanted to be _REALLY_ nitpicky, these should be `$govuk-text-colour` and `$govuk-background-colour` - since they are the versions of black and white that are defined elsewhere - and indeed because it's _not_ #000000, it's #0b0c0c.

Binney (changes requested):

GenderPayGap.WebUI/Views/ActionHubOld/Overview.cshtml
```diff
@@ -0,0 +1,102 @@
...
+<p class="govuk-body-s govuk-!-margin-top-5">
+    <a class="govuk-link print-link" href="/public/assets/pdf/Evidence-based_actions_for_employers.pdf">
+        Print entire guide (PDF)
+    </a>
+</p>
+
+<div class="pdf-preview">
+    <span class="pdf-preview__image">
+        <a href="/public/assets/pdf/Evidence-based_actions_for_employers.pdf">
+            <img style="border:6px solid #dee0e2" width="78" src="/img/what-works.png" alt=""/>
```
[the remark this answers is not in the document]
jamesgriff: I think this image is just for decoration (since the meaning is in the `span` below), so perhaps `alt=""` would be corect
https://en.wikipedia.org/wiki/Alt_attribute#Decorative_images

I suppose really this `<a>` and the next `<a>` should be merged so there's just one link which contains the next and a decorative image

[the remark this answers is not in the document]
dulciejackson: I'm not quite sure which approach to take on this one - nesting the image in the text `<a>` seems weird to me (although that's just a gut reaction which could quite plausibly be entirely wrong)...
@Binney any thoughts?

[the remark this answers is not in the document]
dulciejackson: I tried doing this so the image and text sit inside the same link and I _really_ don't think it's the right approach - it relies on the `<p>` tag containing the download details (PDF, 8.45MB, 12 pages) overlapping with the link area but _not_ being part of the link. It's also really unclear accessibility-wise that the link image and text and the download details are related when the image and text sit in a link together and the details are separately below...not sure what the best method is here, but I think what we've got here with `alt=""` is actually quite good for accessibility compared to the alternatives.

Binney (changes requested):
One comment + the stuff we talked about on Slack regarding the number positions

jamesgriff (approved):

Binney:
Hi @dulciejackson - for the effort of having another conversation about it I have made a few very last tweaks to this:
* Added another `govuk-visually-hidden` class
* The "Effective" page numbers were styled differently to the others because of a missing class
* The indent should be from a padding on the parent ol not a margin on the child li, and I thought it looked a bit big

I'm now happy for this marathon to be merged in 🏃‍♀️

Binney (approved):

PR #282, opened 2020-11-07
GPG-425 Update Admin Pending Registrations

dulciejackson:
[GPG-425 JIRA Ticket](https://technologyprogramme.atlassian.net/browse/GPG-425)

### Context
Admins can view user requests to report for an organisation and/or change the organisation's details. They manually go through these and make sure they seem legitimate.

This work swaps those pages over to the new design system to ensure that the components are as accessible as possible.

### Notes
I've realised that swapping to the new design system here may have broken the validation, since the new GovUkErrorSummary which neatly presents validation errors likely doesn't work without me using
`model.ParseAndValidateParameters(Request, m => m.fieldName)` for each thing I'm trying to validate. Does this warrant a rewrite of the RegisterController.Request as part of this ticket? I'm not entirely sure if validation is even necessary - all the data presented is read only and not being changed by a user...


dulciejackson:
**Update on the GovUkErrorSummary situation:** 
Looking at the code, it seems weird that anything is being validated since it should all be valid to even be displayed on this page! I don't think there's ever a point where we'd expect a validation page from these summary tables, so I've removed the GovUkErrorSummaries as they won't ever get used/the current validation style in the old code means they wouldn't work anyway. 

One day I'll get round to rewriting the horrible controller...

GenderPayGap.WebUI/Views/Register/RequestAccepted.cshtml
```diff
@@ -1,32 +1,39 @@
...
+                </a>
+            }
+            else
+            {
+                <span>@Model.ContactFirstName&nbsp;@Model.ContactLastName</span>
+            }
+            an email confirming they can report on behalf of @(Model.OrganisationName).
+        </p>
+        
+        <button class="govuk-button" href="@Url.Action("PendingRegistrations", "AdminPendingRegistrations")">
```
Binney, line 34: This doesn't work (I think it should be an `<a>`)

GenderPayGap.WebUI/Views/Register/RequestCancelled.cshtml
```diff
@@ -1,31 +1,39 @@
...
+                </a>
+            }
+            else
+            {
+                <span>@Model.ContactFirstName&nbsp;@Model.ContactLastName</span>
+            }
+            an email confirming their registration request for @Model.OrganisationName has been cancelled.
+        </p>
+        
+        <button class="govuk-button" href="@Url.Action("PendingRegistrations", "AdminPendingRegistrations")">
```
Binney, line 35: This also doesn't work

GenderPayGap.WebUI/Views/Register/ReviewAddress.cshtml
```diff
@@ -102,43 +119,128 @@
...
-                <p>
-                    <button type="submit" class="button" name="command" value="approve">Confirm and complete registration</button>
-                </p>
-                <p>
-                    <button type="submit" class="link-button" name="command" value="decline">Reject registration</button>
-                </p>
-            </fieldset>
-        </div>
+            
+        <h2 class="govuk-heading-l">
```
Binney, line 188: The "similar organisations" section is two-thirds width, which I feel is a bit too narrow (it's fine for the above info but makes the table awfully squashed)

GenderPayGap.WebUI/Views/Register/ReviewAddress.cshtml
```diff
@@ -1,35 +1,59 @@
...
+            @Html.HiddenFor(model => model.ManualAddress)
+            
+            @(Html.GovUkErrorSummary())
+
+            <h1 class="govuk-heading-xl">
+                Approve existing employer application - with change of address
             </h1>
+            
             <table class="govuk-table">
                 <caption class="govuk-table__caption govuk-!-margin-bottom-2">Person registering to report and change address</caption>
```
Binney, line 58: I really don't like this caption being the same font-size as the text in the `th`s - I feel it'd make more sense for it to be an h3 (rather than an h2, which would put it on the same level as "Similar organisations"). but I am not an accessibility expert and presumably there was a reason for it to be a `<caption>`?

GenderPayGap.WebUI/Views/Register/ReviewAddress.cshtml
```diff
@@ -102,43 +119,128 @@
...
+                                @{ Dictionary<string, string> references = Model.GetReferences(i); }
+                                @foreach (string key in references.Keys)
+                                {
+                                    if (key.EqualsI("Company No") && references[key].EqualsI(Model.CompanyNumber)
+                                        || key.EqualsI("Charity No") && references[key].EqualsI(Model.CharityNumber)
+                                        || key.EqualsI("Mutual No") && references[key].EqualsI(Model.MutualNumber)
+                                        || key.EqualsI(Model.OtherName) && references[key].EqualsI(Model.OtherValue)
+                                        )
+                                    {
+                                        <span style="color: red; display: block; white-space: nowrap;">@key: @(references[key])</span>
```
Binney, line 220: I do not like inline css. Inline css I do not like

GenderPayGap.WebUI/Views/Register/ReviewRequest.cshtml
```diff
@@ -115,81 +128,116 @@
...
-            <h1 class="heading-medium">
-                Similar organisations
-            </h1>
-            @if (Model.ManualEmployers != null && Model.ManualEmployers.Any())
-            {
-                <table class="govuk-table" summary="A list organisations that match your organisation details">
-                    <caption class="govuk-table__caption govuk-visually-hidden">Customer and Business listing</caption>
-                    <thead class="govuk-table__head">
+            
+        <h2 class="govuk-heading-l">
```
Binney, line 185: Ditto this bit is too narrow at 2/3 imo

Binney (changes requested):
One bug, otherwise looks fine :+1:

GenderPayGap.WebUI/Views/Register/ReviewAddress.cshtml
dulciejackson: Yeah - the caption means that it's read out as a header to the table (and hence if a visually impaired user filters by headings they won't get this one since it only relates to the table content, not to a grouping of other data). This is the "accessible way" recommended by the Design System 😄

GenderPayGap.WebUI/Views/Register/ReviewAddress.cshtml
jamesgriff: I can see the Dulcie copied this from another page.

I'm not sure of the best way to manage CSS that's:
* from old-style pages
* that we want to re-write

We could include it in the new-style Sass/CSS structure, but it would be nice to have a way to make it clear that we don't "approve" of the overall CSS structure of the page, we just want to move some of it out of inline CSS - maybe just some comments?

GenderPayGap.WebUI/Views/Register/ReviewAddress.cshtml
```diff
@@ -89,7 +106,7 @@
                     </tr>
```
jamesgriff, line 106: A couple of lines above here is 
```cshtml
@await Html.PartialAsync("_SicCodeList", Model.SicCodes)
```

We should also convert `_SicCodeList.cshtml` to the new-style classes.
Also, take a look at how we display SIC codes in `ViewOrganisation.cshtml` (lines 188-197) - I think it looks a bit nicer
If we're converting `_SicCodeList.cshtml`, it would be nice to see if we could remove
```cshtml
@inject IDataRepository DataRepo;
```
It's a bit of a bad practice to do database requests from views

GenderPayGap.WebUI/Views/Register/ReviewAddress.cshtml
```diff
@@ -1,35 +1,59 @@
...
+            @Html.HiddenFor(model => model.SectorType)
+            @Html.HiddenFor(model => model.CompanyNumber)
+            @Html.HiddenFor(model => model.CharityNumber)
+            @Html.HiddenFor(model => model.MutualNumber)
+            @Html.HiddenFor(model => model.OtherName)
+            @Html.HiddenFor(model => model.OtherValue)
+            @Html.HiddenFor(model => model.ReviewCode)
+            @Html.HiddenFor(model => model.ManualAddress)
+                        
+            @(Html.GovUkErrorSummary())
```
jamesgriff, line 51: This isn't needed, because:
* this page doesn't have any inputs that could generate validation errors
* (even if the old controller was doing something weird with error messages)
  since these errors are only the ones generated by `viewModel.ParseAndValidate(...)`

GenderPayGap.WebUI/Views/Register/ReviewRequest.cshtml
```diff
@@ -1,195 +1,248 @@
...
+                </tr>
 
+                <tr class="govuk-table__row">
+                    <th class="govuk-table__header" scope="row">
+                        Contact Email
+                    </th>
+                    <td class="govuk-table__cell">
+                        @Model.ContactEmailAddress
+                    </td>
+                </tr>
```
jamesgriff, line 74: Same comments as in other file (e.g. EmailAddress vs ContactEmailAddress)

GenderPayGap.WebUI/Views/Register/RequestAccepted.cshtml
```diff
@@ -1,32 +1,39 @@
-﻿@{
+﻿@model GenderPayGap.WebUI.Models.Register.OrganisationViewModel
+
+@{
     ViewBag.Title = "Registration accepted - Gender pay gap reporting service";
+    Layout = "~/Views/GovUkFrontend/GovUkFrontendLayout.cshtml";
 }
```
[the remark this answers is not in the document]
dulciejackson: Does it make sense to show an ability to navigate back to a page that now technically won't exist? After a request has been confirmed, there won't be a page to review the request, so trying to get back to that page will just show an error.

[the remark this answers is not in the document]
jamesgriff: Ooh, good point - yeh we wouldn't want to create a link that we know wouldn't work.

I suppose we could still have "Admin > Pending Registrations > Request Accepted" - but happy to omit if we have sensible onward links elsewhere on the page

jamesgriff (approved):

GenderPayGap.WebUI/Views/Register/RequestAccepted.cshtml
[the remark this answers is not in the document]
dulciejackson: The "Continue" button returns to the Pending Registrations page so I think that might be ok?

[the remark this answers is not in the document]
jamesgriff: 👍

jamesgriff (approved):

PR #285, opened 2020-11-07
GPG-552: document load test instructions

Binney:
In advance of running some load tests, I've started on some documentation on how to do that. There's also a document in Notion with more detailed plans for when and how we should run load tests, so this is really just the bare minimum technical stuff.

This isn't quite complete - I think we'll need a few changes to the Gatling script because it's a bit old and some of the controllers/pages have been reorganised since the scenario was recorded - but that's to follow.

Would appreciate any feedback on how useful these instructions are :D


dulciejackson (approved):
Looks good!

jamesgriff (approved):

PR #289, opened 2020-11-17
GPG-570 Fix bug in EmailSendingService (re-use of Dictionary which was later modified)

jamesgriff:


dulciejackson (approved):
🎉 🙌 Sorted at last! Thanks for picking up the fix for this! 😄 🎊

Binney (approved):

PR #291, opened 2020-11-20
GPG-588 Fixes screen-reader readout of filter collapses

dulciejackson:
[GPG-588 JIRA Ticket](https://technologyprogramme.atlassian.net/browse/GPG-588)

Fixes the aria-labelling to give meaningful readout on screen-readers. Previously a screen-reader would read out e.g. "Size1 checkbox not checked" (giving no context to what that size might be) whereas now it reads out "Number of employees not provided checkbox not checked" or "500 to 999 employees checkbox not checked" which explains to the user what the checkbox is actually for 😄

### DIFF
--- a/GenderPayGap.WebUI/Views/Viewing/Finder/Parts/Facets/OptionSelect.cshtml
+++ b/GenderPayGap.WebUI/Views/Viewing/Finder/Parts/Facets/OptionSelect.cshtml
@@ -16,7 +16,7 @@
                 @foreach (OptionSelect item in Model.Metadata)
                 {
                     <label for="@(item.Id)" @(item.Disabled ? "disabled" : string.Empty)>
-                        <input name="@(Model.Group)" @(item.Disabled ? "disabled" : string.Empty) value="@(item.Value)" id="@(item.Id)" type="checkbox" @(queryLookup.Contains(item.Value) ? "checked" : "") aria-label="@(item.Id != "Size0" ? item.Label + " employees" : "Number of employees not provided")"/>
+                        <input name="@(Model.Group)" @(item.Disabled ? "disabled" : string.Empty) value="@(item.Value)" id="@(item.Id)" type="checkbox" @(queryLookup.Contains(item.Value) ? "checked" : "")/>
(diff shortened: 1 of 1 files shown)


GenderPayGap.WebUI/Views/Viewing/Finder/Parts/Facets/OptionSelect.cshtml
```diff
@@ -16,7 +16,7 @@
                 @foreach (OptionSelect item in Model.Metadata)
                 {
                     <label for="@(item.Id)" @(item.Disabled ? "disabled" : string.Empty)>
-                        <input name="@(Model.Group)" @(item.Disabled ? "disabled" : string.Empty) value="@(item.Value)" id="@(item.Id)" type="checkbox" @(queryLookup.Contains(item.Value) ? "checked" : "") aria-label="@(item.Id)"/>
+                        <input name="@(Model.Group)" @(item.Disabled ? "disabled" : string.Empty) value="@(item.Value)" id="@(item.Id)" type="checkbox" @(queryLookup.Contains(item.Value) ? "checked" : "") aria-label="@(item.Id != "Size0" ? item.Label + " employees" : "Number of employees not provided")"/>
```
jamesgriff, line 19: I think this OptionSelect.cshtml is used for all the filters (not just number of employees)
So, `item.Label + "employees"` might not always be relevant.

Note: the `item.Label` is there below (on line 20) and I think the aria-label overrides this, so it might just be a case of removing the `aria-label`

dulciejackson: Oh yeah, whoops! 🤦

dulciejackson:
One thing that can't be fixed here is the duplicate readout on a screen-reader - it announces both the clickable area around the checkbox and the checkbox itself with the same label text. I think that'll be hard to fix here so I'm leaving it for now - we'll come back to this page when we redesign it fully :)

jamesgriff (approved):

PR #296, opened 2020-12-09
GPG-571 Fixes malicious excel injection vulnerability

dulciejackson:
Used reference [docs](https://joshclose.github.io/CsvHelper/api/CsvHelper.Configuration/Configuration/) to update CsvHelper config to include the SanitizeForInjection property, which appears to JustWork™️

### DIFF
--- a/GenderPayGap.WebUI/Controllers/Admin/AdminFeedbackController.cs
+++ b/GenderPayGap.WebUI/Controllers/Admin/AdminFeedbackController.cs
@@ -3,6 +3,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using CsvHelper;
+using CsvHelper.Configuration;
 using GenderPayGap.Core;
 using GenderPayGap.Core.Interfaces;
 using GenderPayGap.Database.Models;
@@ -57,7 +58,8 @@ namespace GenderPayGap.WebUI.Controllers
             var memoryStream = new MemoryStream();
             using (var writer = new StreamWriter(memoryStream))
             {
-                using (var csv = new CsvWriter(writer, CultureInfo.CurrentCulture))
+                var config = new CsvConfiguration(CultureInfo.CurrentCulture) { ShouldQuote = (s, context) => true, TrimOptions = TrimOptions.InsideQuotes, SanitizeForInjection = true};
+                using (var csv = new CsvWriter(writer, config))
                 {
                     csv.WriteRecords(feedback.ToList());
                 }
--- a/GenderPayGap.WebUI/Controllers/Admin/AdminFeedbackController.cs
+++ b/GenderPayGap.WebUI/Controllers/Admin/AdminFeedbackController.cs
@@ -58,7 +58,7 @@ namespace GenderPayGap.WebUI.Controllers
             var memoryStream = new MemoryStream();
             using (var writer = new StreamWriter(memoryStream))
             {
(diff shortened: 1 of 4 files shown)


GenderPayGap.WebUI/Models/AdminReferenceData/ReferenceDataHelper.cs
```diff
@@ -72,7 +73,7 @@ internal static class ReferenceDataHelper {
             }
             catch (Exception ex)
             {
-                errorMessage = "The selected file could not be uploaded � try again.";
+                errorMessage = "The selected file could not be uploaded � try again.";
```
Binney, line 76: This registers for me as a change, something to do with your character encoding?

Binney (approved):
Awesome stuff! Have you been able to test it at all?

GenderPayGap.WebUI/Controllers/Admin/AdminFeedbackController.cs
```diff
@@ -57,7 +58,8 @@ public FileContentResult DownloadFeedback(bool nonSpamOnly = false)
             var memoryStream = new MemoryStream();
             using (var writer = new StreamWriter(memoryStream))
             {
-                using (var csv = new CsvWriter(writer, CultureInfo.CurrentCulture))
+                var config = new CsvConfiguration(CultureInfo.CurrentCulture) { ShouldQuote = (s, context) => true, TrimOptions = TrimOptions.InsideQuotes, SanitizeForInjection = true};
+                using (var csv = new CsvWriter(writer, config))
```
dulciejackson, line 62: Pretty much the only important bit - which I overlooked in the first commit because I thought everywhere would use the same sensible config! Turns out the `ToCsv` function which sets the config isn't used for the feedback download, which maybe seems like something to refactor in future (or here, if you think it's worth it). 
Having tested, this now works - excel injection just shows the text of the attempted query instead of doing anything dangerous with it!

GenderPayGap.WebUI/Models/AdminReferenceData/ReferenceDataHelper.cs
dulciejackson: Weird - it doesn't show as a change for me in GitKraken...

dulciejackson: Should be fixed now!

GenderPayGap.WebUI/Controllers/Admin/AdminFeedbackController.cs
Binney: Ahhh great spot!

I'd be in favour of refactoring _all_ the CSV readers to use a standard CsvConfiguration which has SanitizeForInjection set to true. That way future CSV exports don't need to remember to add it in and it just happens automatically. Is that a thing we can do?

(Do we use the other ones eg ShouldQuote, TrimOptions everywhere too? could just inject a consistent configuration to everything)

Binney (approved):

GenderPayGap.WebUI/Controllers/Admin/AdminFeedbackController.cs
dulciejackson: Yeah, agreed that that's probably the best way to go with this. All the CSV readers _don't_ use the same config though - I've just realised that I should probably remove ShouldQuote and TrimOptions from here since they weren't set here originally!

PR #299, opened 2020-12-18
GPG-606 Fixes accessible autocomplete import and requests

dulciejackson:
**Load Testing IV: The Fellowship of the Imports**

Updates the import of `accessible-autocomplete.min.css` in code.
`@import` in an SCSS file combines all imports into one file, whereas `@import` in a CSS file performs an HTTP request for each additional CSS file. This can be fixed by removing the `.css` chunk from the end of the import, which I've done here and fixes the 350+ failing requests in the load tests! We're now at 95% successful requests - I think there'll just need to be a final PR to fix the report submission journey 🎉 

Testing:
- Manually checked that updating the import still works as expected
- Ran the load tests after re-build and confirmed increase and successful requests

### DIFF
--- a/LoadTests/src/test/scala/default/RecordingSimulation.scala
+++ b/LoadTests/src/test/scala/default/RecordingSimulation.scala
@@ -44,10 +44,6 @@ class RecordingSimulation extends Simulation {
 		"DNT" -> "1",
 		"Pragma" -> "no-cache")
 
-	val headers_6 = Map(
-		"content-type" -> "text/css"
-	)
-
 	val searchFeeder = Iterator.continually(Map("searchCriteria1" -> "tes", "searchCriteria2" -> s"test_${Random.nextInt(2 * MAX_NUM_USERS) + 1}"))
 	val registrationFeeder = Iterator.continually(Map("email" -> (Random.alphanumeric.take(20).mkString + "@example.com")))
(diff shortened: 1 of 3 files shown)


LoadTests/src/test/scala/default/RecordingSimulation.scala
```diff
@@ -65,7 +70,10 @@ class RecordingSimulation extends Simulation {
 					.headers(headers_1),
 				http("Load crest")
 					.get("/public/govuk_template/assets/stylesheets/images/govuk-crest-2x.png")
-					.headers(headers_1)))
+					.headers(headers_1),
+				http("Load accessible autocomplete")
+  				.get("/assets/stylesheets/accessible-autocomplete.min.css")
+  				.headers(headers_6)))
```
jamesgriff, line 76: We probably don't need this bit any more (since the CSS file is now included in the main SCSS file)

jamesgriff (approved):

LoadTests/src/test/scala/default/RecordingSimulation.scala
dulciejackson: Good point - I meant to remove this, thanks!

Binney (approved):
🎄

PR #300, opened 2021-01-05
GPG-622 Updates signposting to new guidance

dulciejackson:
[GPG-622 JIRA Ticket](https://technologyprogramme.atlassian.net/browse/GPG-622)

Updates one of the viewing service links to an inaccessible PDF to instead point to the newly released guidance hosted on gov.uk (🎉) 

**Testing:** tested various pages locally after deleting PDF file to check for use elsewhere - also updated links in `ActionHubOld` and `ActionHubNew` so PDF can be removed

**Risk:** low risk of introducing new issues

**Documentation:** introduced new external link as a global so it can be reused across the service - this should act as enough documentation to make new link intentions clear


Binney (approved):
PDF byeeeeeeeeeee

jamesgriff (commented):
I think we should try not to break the PDF link (in case people have it bookmarked).
Here's a few options for not breaking the link:
1. Keep the PDF, but not link to it anywhere
  (and maybe even add it to `robots.txt` so that Google stops indexing it)
    * Pros: link stays working
    * Cons: if you've saved the link, you might keep using the old guidance

2. Remove the PDF, and redirect the user to the new guidance
  We could create a new `RedirectController` or `BrokenLinksController` to handle this (and various redirects we might want in the future)
    * Pros: easy for user
    * Cons: I find it annoying when I have a link (which I know is for one page) and the website sends me to a different page - it makes me feel distrustful of the website

3. Replace the PDF link with a simple page saying that the guidance has moved, but with an option to view the old PDF
  e.g. if you visit `/public/assets/pdf/gender-pay-gap-explained.pdf` you are given an HTML page. It will contain a link to e.g. `/public/assets/pdf/gender-pay-gap-explained-old.pdf` to download the PDF.

Alternatively, we might decide it's find to just show a 404 and let the user figure our where the guidance is - what do you think?

dulciejackson:
> I think we should try not to break the PDF link (in case people have it bookmarked).
@jamesgriff Good point! I barely use bookmarks so I'd completely overlooked that people might have bookmarked the PDF.

I'm not particularly keen on 1. because I think keeping the PDF available implies that it's still relevant - what we really want to do is make it extra-clear that the old guidance shouldn't be used anymore. Plus it's inaccessible, so I don't really want to keep it floating around when we could remove it entirely.

Agreed that 2. could be jarring for users expecting to see a PDF but end up being redirected somewhere else - especially as the place they'll end up isn't on the gender pay gap service at all. Could easily feel like a mistake and lessen trust in other areas of the service. But for 3. again I'm not sure I'd want to include a download link for the outdated PDF - perhaps maybe just the gpg email address in case they need it for any particular reason. 

Personally I'm in favour of either:
- A new page where the old PDF used to live, pointing to the new guidance with the gpg contact email in case they need the old version for some reason
- Removing it altogether and just showing a 404, since the majority of users could probably recover from this kind of error alone - although I don't necessarily think that's a good enough excuse to not make this usable where we can

I reckon it'll probably come down to a bit of a discussion about what's worth the time - Jacob will probably have some opinions and if he thinks time to build a new page is excessive then we'll probably settle for the 404 page.

@Binney any thoughts?

jamesgriff (approved):

PR #303, opened 2021-01-11
GPG-606 Adds new change personal details journey to scenario

dulciejackson:
The old change personal details journey was a single page, so those requests were failing now that the journey has been changed to multiple separate pages. This PR updates the load test scenario to navigate to the Manage Account page, then each sub-page and save some changes to the personal details. 

⚠️ Since this involves changing some personal/login details, the `SET_UP_TEST_ENTITIES` script now needs to be run against the db before each run of the load test scenario.

Test: 
- Ran the `SET_UP_TEST_ENTITIES` script against the db and made a few changes to avoid deletion errors
- Ran the load test scenario and saw the tests passing

Risk: very low

Documentation: clearly named each new step in the load tests so its purpose is clear for future dev work

### DIFF
--- a/LoadTests/src/test/scala/default/RecordingSimulation.scala
+++ b/LoadTests/src/test/scala/default/RecordingSimulation.scala
@@ -569,27 +569,84 @@ class RecordingSimulation extends Simulation {
 				regex("Manage your account")))
 			.pause(PAUSE_MIN_DUR, PAUSE_MAX_DUR)
 
-		// TODO: Update these - change personal details is now multiple pages
-		val visitChangePersonalDetails = exec(http("Visit change personal details")
-			.get("/manage-account/change-details")
+		val visitChangeEmailAddressPage = exec(http("Visit change email address page")
+			.get("/manage-account/change-email")
 			.headers(headers_0)
 			.check(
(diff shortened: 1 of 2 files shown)


LoadTests/src/test/scala/default/RecordingSimulation.scala
```diff
@@ -569,27 +569,84 @@ class RecordingSimulation extends Simulation {
...
+				status.is(200),
+				regex("Saved changes to personal details")))
+  		.pause(PAUSE_MIN_DUR, PAUSE_MAX_DUR)
+
+		val visitChangeContactPreferencesPage = exec(http("Visit change contact preferences page")
+			.get("/manage-account/change-contact-preferences")
+			.headers(headers_0)
+			.check(
+				status.is(200),
+				css("input[name='__RequestVerificationToken']", "value").saveAs("requestVerificationToken"),
```
Binney, line 637: How come this changes?

Binney (approved):
Niceeeee

LoadTests/src/test/scala/default/RecordingSimulation.scala
dulciejackson: I don't think it does, but I wanted to make sure that `requestVerificationToken` has the right value wherever you're coming from on the site, so saved it here for safety!

PR #304, opened 2021-01-12
Gpg 556 open url redirection

AlexJacksonDS:
Updated calls that were Redirect(returnUrl) to LocalRedirect(returnUrl) and the corresponding unit tests.

return Redirect(url) and return new RedirectResult(url) look to be equivalent, so I've changed those using Redirect to return a RedirectResult to simplify the new unit test I've added that searches for instances of ' Redirect(' in the code (similarly to the Html.Raw one). This probably reduces code consistency overall, as we seem to mostly be using Controller methods in returns rather than returning new ActionResult objects, but they are all consistent with each other now i.e. all non-local Redirects are using return new RedirectResult. I did it this way as it makes the condition we are looking for in the new unit test simpler.

### DIFF
--- a/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
+++ b/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using GenderPayGap.WebUI.Tests.TestHelpers;
 using NUnit.Framework;
 
@@ -45,19 +46,22 @@ namespace GenderPayGap.WebUI.Tests.CodeQualityTests
                     continue;
                 }
 
-                string fileText = File.ReadAllText(filePath);
-                // There are a couple of actions on controllers Redirect() that wrap a RedirectToAction,
-                // we want to exclude these from the check as we are only concerned about redirecting with a url parameter
-                if (fileText.Contains(" Redirect(") && !fileText.Contains(" Redirect()"))
+                // We want to pick up all instances of Redirect(url) or new RedirectResult(url) that do not have the disable comment on the previous line.
+                // Ignores those without parameters so we don't pick up any methods defined as Redirect() for example
+                var lines = File.ReadAllLines(filePath);
+                for (var i = 0; i<lines.Length; i++)
                 {
-                    failedFiles.Add(filePathSuffix);
+                    if (Regex.IsMatch(lines[i], @"(new){0,1}\s+Redirect(Result){0,1}\s*\(\S+\)") && !lines[i-1].Contains("//disable:DoNotUseRedirectWithReturnUrls"))
+                    {
+                        failedFiles.Add(filePathSuffix);
+                    }
                 }
             }
 
             // Assert
             if (failedFiles.Any())
             {
-                Assert.Fail($"The following {failedFiles.Count} files contain a call to Redirect with a return url, which is not allowed:\n- {string.Join("\n- ", failedFiles)}\n");
+                Assert.Fail($"The following {failedFiles.Count} files contain a Redirect.\nIf this is to a local url LocalRedirect should be used, otherwise the redirect can be marked to be ignored by this test by adding the \n'//disable:DoNotUseRedirectWithReturnUrls' comment on the preceding line:\n- {string.Join("\n- ", failedFiles.Distinct())}\n");
             }
         }
     }
--- a/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
+++ b/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
@@ -1,8 +1,7 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Reflection;
+using GenderPayGap.WebUI.Tests.TestHelpers;
 using NUnit.Framework;
 
 namespace GenderPayGap.WebUI.Tests.CodeQualityTests
@@ -22,7 +21,7 @@ namespace GenderPayGap.WebUI.Tests.CodeQualityTests
             };
 
             // Arrange
-            string rootCodeFolder = GetRootCodeFolder();
+            string rootCodeFolder = CodeQualityTestHelpers.GetRootCodeFolder();
 
             // Pre-Act Assert (to check we're running the test on the right folder)
             Assert.That(File.Exists($"{rootCodeFolder}\\GenderPayGap.sln"), $"We expect to find a file [GenderPayGap.sln] in the root folder [{rootCodeFolder}]");
@@ -41,7 +40,7 @@ namespace GenderPayGap.WebUI.Tests.CodeQualityTests
                 string filePath = fileInfo.FullName;
                 string filePathSuffix = filePath.Replace(rootCodeFolder, "");
 
-                if (FileIsExcluded(filePathSuffix, excludedFilesAndFolders))
+                if (CodeQualityTestHelpers.FileIsExcluded(filePathSuffix, excludedFilesAndFolders))
                 {
                     continue;
                 }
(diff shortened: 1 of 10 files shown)


GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            foreach (string excludedFilePath in excludedFilesAndFolders)
+            {
+                if (filePathSuffix.StartsWith(excludedFilePath))
+                {
+                    return true;
+                }
+            }
+            return false;
+        }
```
dulciejackson, line 85: These look like duplicates of the methods in `DoNotAllowHtmlDotRawTests.cs` - can we commonise them into a `CodeQualityTestHelpers` file?

dulciejackson (commented):
Looks good - just one small suggestion! 😄

Binney (approved):

dulciejackson (approved):

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                if (CodeQualityTestHelpers.FileIsExcluded(filePathSuffix, excludedFilesAndFolders))
+                {
+                    continue;
+                }
+
+                string fileText = File.ReadAllText(filePath);
+                // There are a couple of actions on controllers Redirect() that wrap a RedirectToAction,
+                // we want to exclude these from the check as we are only concerned about redirecting with a url parameter
+                if (fileText.Contains(" Redirect(") && !fileText.Contains(" Redirect()"))
```
Binney, line 51: I'm not mega keen on this check, I'm concerned e.g. you'll get edge cases with the wrong sort of whitespace at the start (a regex check would be better).

More importantly, I'm not sure this checks the right thing - it enforces future developers use `new RedirectResult` rather than `Redirect`, but what we really want to do is enforce that they have used `LocalRedirect` if possible (and if it's a legitimate external redirect, they can mark it as such and this test won't fail).

Ideally I think this check should flag if
* there's a `Redirect` or `new RedirectResult` which isn't a `LocalRedirect`
* _and_ they haven't stuck a comment on the previous line saying `// disable:DoNotAllowRedirectWithReturnUrlInCode` or something

And then the failure text should inform the user they should either switch to a `LocalRedirect` or stick the comment in there if they understand the risks. WDYT?

Binney (changes requested):

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
AlexJacksonDS: Yeah I'm not super happy about it either and just followed the pattern of the Html.Raw test. Will look into how to do it a different way.

Do you have any particular thoughts on the use of Redirect vs new RedirectResult? If I change the unit test, I can change those so that they are more consistent.

Binney: Good question. Sounds like they're equivalent so security wise I don't mind - but if it makes the test easier to write you could enforce we consistently use one of the two? Up to you!

AlexJacksonDS: Updated to use a regex check to pick up instances of 'Redirect(url)' or 'new RedirectResult(url)' and included the check for a comment on the previous line to ignore an instance.

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
```diff
@@ -45,19 +46,22 @@ public void DoNotAllowRedirectWithReturnUrlInCode()
...
+                        failedFiles.Add(filePathSuffix);
+                    }
                 }
             }
 
             // Assert
             if (failedFiles.Any())
             {
-                Assert.Fail($"The following {failedFiles.Count} files contain a call to Redirect with a return url, which is not allowed:\n- {string.Join("\n- ", failedFiles)}\n");
+                Assert.Fail($"The following {failedFiles.Count} files contain a Redirect.\nIf this is to a local url LocalRedirect should be used, otherwise the redirect can be marked to be ignored by this test by adding the \n'//disable:DoNotUseRedirectWithReturnUrls' comment on the preceding line:\n- {string.Join("\n- ", failedFiles.Distinct())}\n");
```
Binney, line 64: (tiny nit) this line is very long, can you chuck in a line break or two so it's more readable?

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
```diff
@@ -45,19 +46,22 @@ public void DoNotAllowRedirectWithReturnUrlInCode()
...
-                // There are a couple of actions on controllers Redirect() that wrap a RedirectToAction,
-                // we want to exclude these from the check as we are only concerned about redirecting with a url parameter
-                if (fileText.Contains(" Redirect(") && !fileText.Contains(" Redirect()"))
+                // We want to pick up all instances of Redirect(url) or new RedirectResult(url) that do not have the disable comment on the previous line.
+                // Ignores those without parameters so we don't pick up any methods defined as Redirect() for example
+                var lines = File.ReadAllLines(filePath);
+                for (var i = 0; i<lines.Length; i++)
                 {
-                    failedFiles.Add(filePathSuffix);
+                    if (Regex.IsMatch(lines[i], @"(new){0,1}\s+Redirect(Result){0,1}\s*\(\S+\)") && !lines[i-1].Contains("//disable:DoNotUseRedirectWithReturnUrls"))
```
Binney, line 54: I think it's generally better/faster to use a compiled regex here - https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices so you'd pull out a `var redirectResult = new Regex(....)` outside of the loop/test method and use that

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
```diff
@@ -45,19 +46,22 @@ public void DoNotAllowRedirectWithReturnUrlInCode()
...
                 }
 
-                string fileText = File.ReadAllText(filePath);
-                // There are a couple of actions on controllers Redirect() that wrap a RedirectToAction,
-                // we want to exclude these from the check as we are only concerned about redirecting with a url parameter
-                if (fileText.Contains(" Redirect(") && !fileText.Contains(" Redirect()"))
+                // We want to pick up all instances of Redirect(url) or new RedirectResult(url) that do not have the disable comment on the previous line.
+                // Ignores those without parameters so we don't pick up any methods defined as Redirect() for example
+                var lines = File.ReadAllLines(filePath);
+                for (var i = 0; i<lines.Length; i++)
```
Binney, line 52: I'm a bit worried about the speed/efficiency impact of this - C# will have all sorts of weird internal optimisations for the regex-only solution, but manually looping through the lines will lose it.

Can you think of a regex that matches only the right cases? Also, how long does the test take to run on your machine? (It's not production code so I'm not too bothered if a regex solution isn't obvious)

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
Binney, line 54: you can use `?` instead of `{0, 1}`

Binney, line 54: Usual boring reminder you'll get an `IndexOutOfArrayException` if you've got a redirect on the first line of the file (unlikely I know...)

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
Binney: Alternatively, how about 
* First pass: does it match the redirect regex at all?
* If it does, do a second pass - go through all the lines in just that one file and do this check?

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/CodeQualityTests/DoNotUseRedirectWithReturnUrls.cs
```diff
@@ -45,19 +46,22 @@ public void DoNotAllowRedirectWithReturnUrlInCode()
                     continue;
                 }
 
-                string fileText = File.ReadAllText(filePath);
-                // There are a couple of actions on controllers Redirect() that wrap a RedirectToAction,
-                // we want to exclude these from the check as we are only concerned about redirecting with a url parameter
-                if (fileText.Contains(" Redirect(") && !fileText.Contains(" Redirect()"))
+                // We want to pick up all instances of Redirect(url) or new RedirectResult(url) that do not have the disable comment on the previous line.
+                // Ignores those without parameters so we don't pick up any methods defined as Redirect() for example
```
Binney, line 50: +1 to this comment

Binney (changes requested):
Nice! Got some suggestions, thoughts welcome (imo this version is basically fine but we might be able to improve even further)

AlexJacksonDS:
Changes made to fix the long Assert line and replace {0,1} with ? in the regex. 
Regarding the performance, I tried moving the regex into an object outside the loop as suggested, but it made no difference to the run time of the unit test, remaining at a random ~180-230ms.

Binney (approved):
Looks good to me! :+1:

PR #305, opened 2021-01-12
GPG-471 Search Button Contrast

dulciejackson:
[GPG-471 JIRA Ticket](https://technologyprogramme.atlassian.net/browse/GPG-471)

Updates search icon to use white 50px x 50px SVG icon instead of black and white PNG variants. This also fixes the accessibility problem with the pixellated search icon. Unfortunately, the SVG must be inline for the `fill` attribute to work as required, but given the small file this shouldn't be too much of a problem.

Testing: tried using the button, including hovering and focus. Also tried using variants of the Windows High Contrast mode, and confirmed that SVG colour changes to match contrast mode.

Risk: very low - checked for other usages of the search classes (there were none) so unlikely to affect other areas of the service

Documentation: none needed - CSS classes have clear names to improve maintainability


GenderPayGap.WebUI/Views/Viewing/Launchpad/Parts/SearchByOrganisation.cshtml
```diff
@@ -9,7 +9,24 @@
                     <!-- The component is created by the `generateOrganisationSearchbar` function in Index.cshtml -->
                 </div>
                 <div class="gov-uk-l-searchbar__cell">
-                    <button class="gov-uk-c-searchbar__submit gov-uk-c-searchbar__submit--black" id="NextStep" type="submit">Search</button>
+                    <button class="govuk-c-searchbar__submit govuk-c-searchbar__submit--black" id="NextStep" type="submit">
```
Binney, line 12: So I'm kicking myself that there must be a better way to do this but I'm not sure what it is. Some other random people on the internet ([1](https://vega.rd.no/2018/02/cool-stuff-with-asp.net-core-tag-helpers-and-font-awesome/), [2](https://gist.github.com/NickCraver/2d8c49374059792830e278fe406b2d79)) have built helpers to embed an SVG as a little "include" partial but they're quite heavy duty (and, hilariously, involve `Html.Raw()` which we are trying to remove).

Can you think of a way to make this reusable or shall we just leave it inline?

Binney (approved):
Looks good to me! I'm a bit sad about the un-reusability of the SVG but if this is the only icon in the website it's not worth making generic

GenderPayGap.WebUI/Views/Viewing/Launchpad/Parts/SearchByOrganisation.cshtml
dulciejackson: Yeah, it's a massive pain - I did a bit of a dive into how inline vs imported SVGs work as I was doing this ticket, and still haven't been able to track down a method that allows this to be reusable whilst also letting us style it as we want 😢 

Given that this SVG is only used here (and in future possibly on the admin-side search), I doubt it's worth investing the time in building a helper for this - perhaps we can keep an eye on how many places we want to do things like this, and rethink if we find we're having to use inline SVGs a lot?

Binney: Works for me!

PR #306, opened 2021-01-13
GPG-598 Orders returns by ReturnStatus

dulciejackson:
[GPG-598 JIRA Ticket](https://technologyprogramme.atlassian.net/browse/GPG-598)

Smallest PR in the world - just changing the ordering so Returns are ordered by Status and then Modified date. Historically, the Modified date was also used to determine late flags, so the active Return could not be the first one shown - ordering by Status first ensures that this isn't the case.

Test: checked business logic with Raam, and checked with an old return that the rows now show in correct order

Risk: low

Documentation: none needed - the ordering is clear in the code and ordering by Status makes sense without need for further justification


Binney (approved):
I love it when code is straightforward

PR #308, opened 2021-01-13
GPG-573 Change password reset code to Crypto string

AlexJacksonDS:
Changed the SendPasswordResetEmail method to use the Crypto.GetSalt method to produce a base64 string instead of Guid.NewGuid.


dulciejackson (approved):
Good job spotting those typos! 👀

Binney (approved):

PR #309, opened 2021-01-13
GPG-581 Add middleware to add no-cache to headers

AlexJacksonDS:
Added middleware to set the Cache-Control header to no-cache for all pages. Css etc. still appear to be cached fine locally after this change. If we decide that we don't want to have the no-cache header for everything, this will probably have to be done differently, probably via ResponseCacheAttributes on the relevant controllers.

### DIFF
--- a/GenderPayGap.WebUI/Startup.cs
+++ b/GenderPayGap.WebUI/Startup.cs
@@ -373,7 +373,8 @@ namespace GenderPayGap.WebUI
                 app.UseMiddleware<BasicAuthMiddleware>();
             }
 
-            app.UseMiddleware<NoCacheMiddleware>();
+            // Prevent caching of html responses as they may contain secure info - GPG-581
+            app.UseMiddleware<NoHtmlCachingMiddleware>();
 
             app.UseEndpoints(endpoints =>
             {
--- a/GenderPayGap.WebUI/Startup.cs
+++ b/GenderPayGap.WebUI/Startup.cs
@@ -373,6 +373,8 @@ namespace GenderPayGap.WebUI
                 app.UseMiddleware<BasicAuthMiddleware>();
             }
 
+            app.UseMiddleware<NoCacheMiddleware>();
+
             app.UseEndpoints(endpoints =>
             {
                 endpoints.MapControllers();
--- a/GenderPayGap.WebUI/Helpers/NoCacheMiddleware.cs
+++ b/GenderPayGap.WebUI/Helpers/NoCacheMiddleware.cs
@@ -3,6 +3,11 @@ using Microsoft.AspNetCore.Http;
 
 namespace GenderPayGap.WebUI.Helpers
 {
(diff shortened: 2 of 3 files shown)


dulciejackson (commented):
Looks good! 
Would be helpful to add a couple of comments to these files just so we're aware what this does (you can reference the ticket number to make it clear when the problem was noticed), and what the `no-cache` option does in this context 😄

AlexJacksonDS:
I've changed it to be using 'no-store' instead as that is the header value that will actually provide the desired behaviour after looking at the documentation when adding the comment. See comments on  [GPG-581](https://technologyprogramme.atlassian.net/browse/GPG-581) for more info.
I've also renamed the new class since it is no longer adding no-cache

dulciejackson:
Having done a bit of reading about `no-cache` vs `no-store`, I think `no-cache` is generally preferred since `no-store` can cause significant speed and usability issues (e.g. back buttons not working). 
A few things I've read have suggested using `no-cache` everywhere, and `no-store` in specific places where sensitive data could be available - @Binney do you think that sounds like a reasonable approach.

Also, just to make things difficult, IE 5-8 error when the `no-cache` or `no-store` headers are used ([1](https://support.microsoft.com/en-us/help/812935/internet-explorer-cannot-download-error-message-when-you-use-an-https)) - we agreed not to support IE < 8, but since this is also present on IE 8 we probably need to try and find a way to stop these errors.

GenderPayGap.WebUI/Startup.cs
```diff
@@ -373,6 +373,9 @@ public void Configure(IApplicationBuilder app, IApplicationLifetime lifetime)
                 app.UseMiddleware<BasicAuthMiddleware>();
             }
 
+            // Prevent caching of html responses as they may contain secure info - GPG-581
```
Binney, line 376: :+1:

Binney (approved):
That IE issue is bonkers!! We should definitely test it with browserstack once it's on Dev - it does appear to be something to do with PDFs/file downloads though, so I'm tentatively optimistic our web pages will be fine.

I don't know the downsides of `no-store`, I'm afraid. I don't think we should be writing middleware more complex than this to send different headers under different circumstances (it's just not worth the trouble) so I think the choice is
* no-cache, max-age: 0
* no-store

and I'll let Dulcie pick between those!

dulciejackson:
Thanks Sarah! 

I think we should be using `no-store` here after all, since we're concerned about both:
- a malicious entity spying on something cached, and using it to validate themselves
_and_
- a malicious entity spying on something cached, extracting sensitive data, and using it elsewhere

`no-cache` will only prevent the first of these two, so I think `no-store` is the best approach here. As Sarah said, we should test this through BrowserStack once it's on Dev so we're sure it doesn't cause any browser-specific issues!

dulciejackson (approved):

PR #310, opened 2021-01-14
GPG-583 Set HttpOnly and Secure on cookies

AlexJacksonDS:
Updated the AspNetCore.Session cookie to be HttpOnly and the AspNetCore.Antiforgery.xxxxx and AspNetCore.Cookies to have the Secure flag set.

cookie_settings and seen_cookie_message are left as before since the JS needs to interact with them. These show for me with HttpOnly false and Secure true running locally.

### DIFF
--- a/GenderPayGap.WebUI/Startup.cs
+++ b/GenderPayGap.WebUI/Startup.cs
@@ -101,8 +101,12 @@ namespace GenderPayGap.WebUI
                 mvcBuilder.AddRazorRuntimeCompilation();
             }
 
-            //Add antiforgery token by default to forms
-            services.AddAntiforgery();
+            //Add antiforgery token by default to forms making sure the Secure flag is always set
+            services.AddAntiforgery(
+                options =>
+                {
+                    options.Cookie.SecurePolicy = CookieSecurePolicy.Always;
+                });
 
             //Add services needed for sessions
             services.AddSession(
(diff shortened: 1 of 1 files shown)


GenderPayGap.WebUI/Startup.cs
```diff
@@ -101,16 +101,20 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
...
+                    options.Cookie.SecurePolicy = CookieSecurePolicy.Always;
+                });
 
             //Add services needed for sessions
             services.AddSession(
                 o =>
                 {
                     o.Cookie.IsEssential = true; //This is required otherwise session will not load
                     o.Cookie.SecurePolicy = CookieSecurePolicy.Always; //Equivalent to <httpCookies requireSSL="true" /> from Web.Config
-                    o.Cookie.HttpOnly = false; //Always use https cookies
```
Binney, line 113: I love how the previous developer misunderstood what this does

GenderPayGap.WebUI/Startup.cs
```diff
@@ -101,16 +101,20 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
...
+                    o.Cookie.HttpOnly = true; //Session cookie should not be accessible by client-side scripts
```
Binney, line 117: :+1:

Binney (approved):
Nice :)

dulciejackson (approved):

PR #311, opened 2021-01-15
GPG-616 Change the Manage Account/Organisation aria roles to tabs

AlexJacksonDS:
This seems to be the correct setup for tabs after playing around with roles/labels in different places and a bit of googling. The presentation role to hide the list items from the reader and declaring the links as tabs.

### DIFF
--- a/GenderPayGap.WebUI/Views/Components/Navigation/AccountTabs.cshtml
+++ b/GenderPayGap.WebUI/Views/Components/Navigation/AccountTabs.cshtml
@@ -10,13 +10,13 @@
     <div class="govuk-grid-column-full">
 
         <nav class="account-tabs">
-            <ul>
+            <ul role="tablist">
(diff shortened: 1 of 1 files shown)


dulciejackson (approved):
Looks great 😄 
For reference, I think in an ideal world the `tab` role would be paired with `tabpanel` roles for the content "inside" each tab [(1)](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role). We can't do this here because these tabs don't actually function like tabs (they're basically just links positioned like tabs) so there's still an ongoing accessibility issue with this page, but marking them up properly will help visually impaired users navigate the site a lot better! I'm hopeful that in future we'll be able to properly give this some thought to improve the accessibility of this page in particular 🤞

Binney (approved):

PR #312, opened 2021-01-18
GPG-624 Fix nullable object access error  in ViewOrganisation view

AlexJacksonDS:
Turns out it was that incorrect 'else if' statement I noticed on Friday that was causing an error when calling .Value on a nullable date since there is no PINConfirmedDate while the user is still pending

### DIFF
--- a/GenderPayGap.WebUI/Views/Admin/ViewOrganisation.cshtml
+++ b/GenderPayGap.WebUI/Views/Admin/ViewOrganisation.cshtml
@@ -297,7 +297,7 @@ else
                 {
                     userStatus = "AWAITING_PIN";
                 }
-                else if (userOrganisation.IsAwaitingActivationPIN())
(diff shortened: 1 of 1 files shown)


GenderPayGap.WebUI/Views/Admin/ViewOrganisation.cshtml
```diff
@@ -297,7 +297,7 @@ else
                 {
                     userStatus = "AWAITING_PIN";
                 }
-                else if (userOrganisation.IsAwaitingActivationPIN())
+                else if (userOrganisation.IsAwaitingRegistrationApproval())
```
Binney, line 300: I love a nice cup of NullPointerException in the morning

Binney (approved):

dulciejackson (approved):
Great spot! :)

PR #313, opened 2021-01-19
Change the health check to output a json status

AlexJacksonDS:
Figured I might as well just get your opinions before swapping the output to this


AlexJacksonDS:
Merged to fix dev deployment

PR #314, opened 2021-01-21
Gpg 629 amend guidance template

AlexJacksonDS:
I've pulled BaseController out of the split controller since nothing was actually using anything from it. Both the 'Old' (current) and 'New' still work fine after doing this. It might be worth changing the names here if we are going to be working with the old version and the 'new' may never get used.


GenderPayGap.WebUI/Controllers/ActionHubController.New.cs
```diff
@@ -1,31 +1,14 @@
...
-using GenderPayGap.Extensions.AspNetCore;
-using GenderPayGap.WebUI.Classes;
 using Microsoft.AspNetCore.Mvc;
 
 namespace GenderPayGap.WebUI.Controllers
 {
 
     [Route("actions-to-close-the-gap")]
-    public partial class ActionHubController : BaseController
+    public partial class ActionHubController : Controller
```
dulciejackson, line 10: 🎉

GenderPayGap.WebUI/Views/ActionHubOld/Index.cshtml
```diff
@@ -41,46 +86,57 @@
...
+                    <a class="govuk-link govuk-body-s" href="/public/assets/pdf/action-plan-guidance.pdf" rel="download">
+                        Four steps to developing a gender pay gap action plan
+                        <span class="govuk-visually-hidden"> (pdf)</span>
                     </a>
                 </li>
                 <li>
-                    <a class="govuk-link govuk-body-s" href="/public/assets/pdf/action-plan-guidance.pdf">
-                        Four steps to developing a gender pay gap action plan (PDF)
+                    <a class="govuk-link govuk-body-s" href="@Url.Action("EightWaysOverview", "EightWaysGuidance")">
+                        Eight ways to understand your oganisation's gender pay gap'
```
dulciejackson, line 101: Should the `'` at the end of this line be there?

GenderPayGap.WebUI/Views/ActionHubOld/Index.cshtml
```diff
@@ -41,46 +86,57 @@
...
                     </a>
                 </li>
                 <li>
-                    <a class="govuk-link govuk-body-s" href="/public/assets/pdf/gender-pay-gap-explained.pdf">
-                        Gender pay gap explained (PDF)
+                    <a class="govuk-link govuk-body-s" href="@Url.Action("SearchResults", "Viewing")">
+                        Find and compare employer’s gender pay gaps
```
dulciejackson, line 106: Grammatically, I think this maybe should be `employers' gender pay gaps` since I read this as talking about multiple employers. May be reading it wrong, but probably best to check (probably with Claire) in case!

dulciejackson (commented):
Looks good! Just a couple of things 😄

GenderPayGap.WebUI/Views/ActionHubOld/Index.cshtml
AlexJacksonDS: Nope, it's a typo, I'll sort that

dulciejackson (approved):
Awesome! 😄

Binney (approved):
Agreed with Dulcie on the apostrophe thing but sounds like we're going with "employer" anyway, so all looks good to me :+1:

AlexJacksonDS:
Having talked to Dulcie, I'm actually going to remove all the ActionHubNew stuff as it is unlikely to ever be used (and is confusing). Will make those for re-review.

AlexJacksonDS:
Removed ActionHubController.New and corresponding views, removing the Old from the remaining ActionHub files. Moved around and renamed some of the views to make them more consistent with the other similarly structured guidance pages associated with EightWaysGuidanceController

dulciejackson (approved):
This all looks great 😄

Binney (approved):

PR #315, opened 2021-01-22
Updates dependencies to fix dependabot conflict

dulciejackson:
Dependabot was reporting a conflicting dependency update. `yargs-parser` is a transitive dependency of `node-sass`, but there is a security vulnerability in the version of `yargs-parser` currently being used by our version of `node-sass`.

I've manually updated `node-sass`, which has allowed `yargs-parser` to be updated to a version with the security patch. Whilst I was at it, I updated a couple of other dependencies.

Testing: I've been through the site locally after these dependency changes were made, and everything I tried seems to work as I'd expect. We'll want to do some more thorough testing on dev to be completely sure before we release this.

Risk: moderate - we want to be totally sure that these upgrades don't deprecate any functionality we rely on. We should comprehensively test the service on dev (and preprod) to ensure everything is stable.

Documentation: none needed

### DIFF
--- a/GenderPayGap.WebUI/package-lock.json
+++ b/GenderPayGap.WebUI/package-lock.json
@@ -8,9 +8,9 @@
       "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
     },
     "ajv": {
-      "version": "6.12.4",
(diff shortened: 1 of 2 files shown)


Binney (approved):
As long as testing looks fine these all look good to me, might be good to double check the changelogs on the govuk-frontend package in case there's anything we should know about? but the other two not worth it

dulciejackson:
> 
> 
> might be good to double check the changelogs on the govuk-frontend package in case there's anything we should know about?

I've already done this! Nothing particularly exciting - a couple of accessibility fixes look like the biggest updates so we should be fine 😄

PR #316, opened 2021-01-26
GPG-643 Add script to deploy IP denylist app

AlexJacksonDS:
Initial state of it, not ready to merge.

### DIFF
--- a/Infrastructure/gpg-ipdeny/README.md
+++ b/Infrastructure/gpg-ipdeny/README.md
@@ -1,0 +1,44 @@
+# gpg-ipdeny
+
+gpg-ipdeny is an IP denylisting nginx application based upon the IP allowing version found here: [GOV.UK PaaS IP authentication route service](https://github.com/alphagov/paas-ip-authentication-route-service)
+
+It is designed to allow access to the Gender Pay Gap Service to all IP addresses except those specifically set, returning a 403 response to those banned IPs.
+
+## Deployment
+
+Since the list of banned IPs should not update particularly frequently, the application is deployed manually via the Deploy.sh script.
+
+The script relies up having the command line processor [jq](https://stedolan.github.io/jq/) available. 
+
+When using gitbash for Windows this can simply be added by running the following command, it should also be possible to add it to the system path:
+```
+curl -L -o /usr/bin/jq.exe https://github.com/stedolan/jq/releases/latest/download/jq-win64.exe
+```
+
+Before running the deployment script you must be logged into GOV.UK PaaS using the Cloud Foundry CLI.
+
+The deployment script takes the following parameters:
+* -a PROTECTED_APP_NAME The application to protect
+* -e PROTECTED_APP_SPACE_NAME The name of the space the apps are in
+* -i DENIED_IPS A comma separated list of IP addresses to be banned
+* -r ROUTE_SERVICE_APP_NAME The name of the service app to create/update
+* -s ROUTE_SERVICE_NAME The name of the service to create/update
+
+For example to set up ip denylisting on the dev space the command could be run (using the gitbash window set up with jq earlier) as 
+```
+./deploy.sh -e gpg-dev -r gpg-dev-ipdeny -s gpg-dev-ipdeny-service -a gender-pay-gap-dev -i "1.2.3.4,2.3.4.5"
+```
+
+The comma separated list of banned IPs can be found in Zoho, when updating this list, newlines should be avoided as they can cause syntax errors within the nginx.conf. This can be seen by an error when deploying, and then if you try and run
+```
+cf start ROUTE_SERVICE_APP_NAME
+```
+an error similar to the following will display
+```
+2021/01/25 15:25:07 [emerg] 29#0: unknown directive "
+   " in /tmp/conf410448399/nginx.conf:73
+   nginx: configuration file /tmp/conf410448399/nginx.conf test failed
+          **ERROR** Could not validate nginx.conf: nginx.conf contains syntax errors: exit status 1
+   Failed to compile droplet: Failed to run all supply scripts: exit status 14
+```
+By checking the environment variables that have been set on the environment, it should be possible to tell where the issue is in the provided list of IPs
--- a/Infrastructure/gpg-ipdeny/README.md
+++ b/Infrastructure/gpg-ipdeny/README.md
@@ -20,25 +20,16 @@ Before running the deployment script you must be logged into GOV.UK PaaS using t
 The deployment script takes the following parameters:
 * -a PROTECTED_APP_NAME The application to protect
 * -e PROTECTED_APP_SPACE_NAME The name of the space the apps are in
-* -i DENIED_IPS A comma separated list of IP addresses to be banned
+* -f DENIED_IPS_FILENAME The filename containing the list of IPs to block
 * -r ROUTE_SERVICE_APP_NAME The name of the service app to create/update
 * -s ROUTE_SERVICE_NAME The name of the service to create/update
 
-For example to set up ip denylisting on the dev space the command could be run (using the gitbash window set up with jq earlier) as 
+For example to set up ip denylisting on the dev space, if you've created/placed the denylist file in the same directory as the deployment script, the command could be run (using the gitbash window set up with jq earlier) as 
 ```
-./deploy.sh -e gpg-dev -r gpg-dev-ipdeny -s gpg-dev-ipdeny-service -a gender-pay-gap-dev -i "1.2.3.4,2.3.4.5"
+./deploy.sh -e gpg-dev -r gpg-dev-ipdeny -s gpg-dev-ipdeny-service -a gender-pay-gap-dev -f "GPG IP Denylist.txt"
 ```
 
-The comma separated list of banned IPs can be found in Zoho, when updating this list, newlines should be avoided as they can cause syntax errors within the nginx.conf. This can be seen by an error when deploying, and then if you try and run
-```
-cf start ROUTE_SERVICE_APP_NAME
-```
-an error similar to the following will display
-```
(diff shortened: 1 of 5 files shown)


AlexJacksonDS:
Updated to add the IPs as an input parameter as I couldn't get the config vcap way to work. Added a readme which hopefully explains how to use it sufficiently. Should all be fine now

Infrastructure/gpg-ipdeny/manifest.yml
```diff
@@ -0,0 +1,15 @@
+---
+
+applications:
+  - name: gpg-ipdeny
+    instances: 1
+    memory: 256M
```
Binney, line 6: We will want to think about this number (and this app generally) during the load test - as all our traffic will be going through it. Not one for this review but for a few weeks' time

Infrastructure/gpg-ipdeny/nginx.conf
```diff
@@ -0,0 +1,107 @@
...
+                         ' "time": "$time_local", '
+                         ' "request": "$request", '
+                         ' "status": $status, '
+                         ' "size": $body_bytes_sent, '
+                         ' "referer": "$http_referer", '
+                         ' "userAgent": "$http_user_agent", '
+                         ' "requestTime": $request_time, '
+                         ' "httpHost": "$http_host"}';
+
+  access_log /dev/stdout access_json;
```
Binney, line 25: Looks like these will just sit around on the instance and we don't send them anywhere? While I don't think it adds a lot to send the whole thing to logit, I would be very interested in stats on how many users we are denying access to - is that something we can pipe to another log stream and consider uploading to the cloud along with all the app server logs?

Infrastructure/gpg-ipdeny/nginx.conf
```diff
@@ -0,0 +1,107 @@
...
+
+      if ($http_x_cf_forwarded_url ~* ^(https?\:\/\/)(.*?)\/(.*)$) {
+        set $cf_forwarded_host $2;
+        set $cf_forwarded_uri /$3;
+      }
+
+      proxy_http_version 1.1;
+      proxy_ssl_server_name on;
+      proxy_ssl_protocols TLSv1.2;
+      proxy_set_header Connection "";
```
Binney, line 97: Any other headers we're expecting to need to set here? e.g. application login token? (Is there a way to make it pass everything/is that desirable?)

Infrastructure/gpg-ipdeny/deploy.sh
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if ! command -v jq >/dev/null; then
+  echo "Must have JQ installed"
+  exit 1
+fi
+
+while getopts ":a:e:i:r:s:" opt; do
```
Binney, line 9: Niceeeeeeeee

Infrastructure/gpg-ipdeny/README.md
```diff
@@ -0,0 +1,44 @@
...
+* -i DENIED_IPS A comma separated list of IP addresses to be banned
+* -r ROUTE_SERVICE_APP_NAME The name of the service app to create/update
+* -s ROUTE_SERVICE_NAME The name of the service to create/update
+
+For example to set up ip denylisting on the dev space the command could be run (using the gitbash window set up with jq earlier) as 
+```
+./deploy.sh -e gpg-dev -r gpg-dev-ipdeny -s gpg-dev-ipdeny-service -a gender-pay-gap-dev -i "1.2.3.4,2.3.4.5"
+```
+
+The comma separated list of banned IPs can be found in Zoho, when updating this list, newlines should be avoided as they can cause syntax errors within the nginx.conf. This can be seen by an error when deploying, and then if you try and run
```
Binney, line 32: The list in Zoho is quite long and I can see it being quite hard to maintain. With that in mind I wonder how much extra work it would be to use a file here - e.g.
* Argument is a filename
* File is basically "exactly what you found in Zoho" (i.e. what's in Zoho would need editing to remove comments)
* File comprises one entry per line, meaning it's easier to read/maintain
* Script replaces newlines with commas

thoughts?

Infrastructure/gpg-ipdeny/README.md
```diff
@@ -0,0 +1,44 @@
+# gpg-ipdeny
```
Binney, line 1: So the other place that could use some documentation is... there's a script at `Infrastructure/CreateEnvironment.sh` (and corresponding `DeleteEnvironment.sh` which could use updating in line with this. @dulciejackson will be putting them to the test when she builds the load test environment, so they don't necessarily need to be tested but at least a rough guess at the correct commands - which will error if it doesn't work, so we don't forget to fix it.

Binney (approved):
Lovely stuff :+1: Main thoughts are the maintainability of the denylist file, and how we go about testing this in preprod. In particular you should liaise with Dulcie around what this extra infrastructure means for the load test. Looking forward to getting it up and running!

Infrastructure/gpg-ipdeny/README.md
AlexJacksonDS: Agreed that it would be better to do something like that, would just need to figure out how

AlexJacksonDS:
Updated to take a filename parameter instead of a comma separated list of IP addresses (also updated the file in zoho to match)
Updated the nginx.conf so that headers are preserved in responses rather than being changed (particularly cache-control)

Infrastructure/gpg-ipdeny/nginx.conf
```diff
@@ -90,7 +90,7 @@ http {
         set $cf_forwarded_host $2;
         set $cf_forwarded_uri /$3;
       }
-
+
```
Binney, line 93: Some weird whitespace here

Infrastructure/gpg-ipdeny/nginx.conf
```diff
@@ -101,6 +101,12 @@ http {
       # Use XX-CF-APP-INSTANCE on the original request if you wish to target an instance
       proxy_set_header X-CF-APP-INSTANCE $http_xx_cf_app_instance;
 
+      # Setting this allows the asp.net app to set the cache-control response headers
+      expires off;
+
+      # Removes the Server: nginx header from the response
```
Binney, line 107: Niceeeee

Infrastructure/gpg-ipdeny/README.md
```diff
@@ -20,25 +20,16 @@ Before running the deployment script you must be logged into GOV.UK PaaS using t
...
-* -i DENIED_IPS A comma separated list of IP addresses to be banned
+* -f DENIED_IPS_FILENAME The filename containing the list of IPs to block
 * -r ROUTE_SERVICE_APP_NAME The name of the service app to create/update
 * -s ROUTE_SERVICE_NAME The name of the service to create/update
 
-For example to set up ip denylisting on the dev space the command could be run (using the gitbash window set up with jq earlier) as 
+For example to set up ip denylisting on the dev space, if you've created/placed the denylist file in the same directory as the deployment script, the command could be run (using the gitbash window set up with jq earlier) as 
 ```
-./deploy.sh -e gpg-dev -r gpg-dev-ipdeny -s gpg-dev-ipdeny-service -a gender-pay-gap-dev -i "1.2.3.4,2.3.4.5"
+./deploy.sh -e gpg-dev -r gpg-dev-ipdeny -s gpg-dev-ipdeny-service -a gender-pay-gap-dev -f "GPG IP Denylist.txt"
```
Binney, line 29: I'm not a big fan of spaces in filenames but not massively fussed about changing it

Infrastructure/gpg-ipdeny/deploy.sh
```diff
@@ -43,16 +43,16 @@ if [ -z "${PROTECTED_APP_SPACE_NAME+set}" ]; then
...
   exit 1
 fi
 
-IFS="," read -ra IPS <<< "$DENIED_IPS"
+readarray -t IPS < "$DENIED_IPS_FILENAME"
 
 NGINX_DENY_STATEMENTS=""
 for addr in "${IPS[@]}";
-  do NGINX_DENY_STATEMENTS="$NGINX_DENY_STATEMENTS deny $addr;"; true;
+  do NGINX_DENY_STATEMENTS="$NGINX_DENY_STATEMENTS deny ${addr//[$'\r']};"; true;
```
Binney, line 55: Oh yeah I like this way more than commas

Binney: Is it worth windows-proofing? will you get any line ending issues?

Binney (approved):
Aceeee big fan :+1: 

How much testing have you been able to do with the headers - is there anything else we think it is overwriting/blocking? Or is it just the cache and server headers?

Infrastructure/gpg-ipdeny/deploy.sh
AlexJacksonDS: I did have some issues with the line ending hence the ${addr//[$'\r']}, I think the readarray was just splitting on and stripping the \n of the \r\n in the text file. If we find we do need to strip other stuff it should just be a case of adding them here

AlexJacksonDS:
> How much testing have you been able to do with the headers - is there anything else we think it is overwriting/blocking? Or is it just the cache and server headers?

Headers all look the same on dev and preprod now (a few slightly different values for things like content-length) and everything seems to be working fine, e.g. it hasn't re-broken the secure cookies by unsetting the x-forwarded-proto headers. I haven't checked every page though

Infrastructure/gpg-ipdeny/deploy.sh
Binney: That works for me, as long as there isn't something easy we should be doing to make it a bit more resilient

dulciejackson (approved):
Looks great 😄 
I'll chat with you about anything we'll need to configure with this once I get round to running a load test!

PR #317, opened 2021-01-28
GPG-622 Retrieve broken PDF from source control

AlexJacksonDS:
Retrieving broken pdf from source control


Binney:
We shouldn't be linking to this PDF anywhere - we're trying to move away from PDFs generally. Where is it linked from and can we link to the appropriate guidance page instead?

(I'm not against keeping the PDF undeleted for a bit longer in case anyone has it bookmarked, but we certainly shouldn't be directing people to it)

AlexJacksonDS:
It was linked from the homepage and the guidance related content sections, which are being changed in GPG-629 to no longer link to it

Binney:
Nice, I'm quite sad we deleted it before the link was removed, but either way merging this and GPG-629 in for the next release looks like the right way to go :+1:

Binney (approved):

AlexJacksonDS:
> Nice, I'm quite sad we deleted it before the link was removed, but either way merging this and GPG-629 in for the next release looks like the right way to go 👍

It wasn't actually 'deleted' as far as main is concerned, it was removed then added back as it was still needed in 2 commits for 622, but seems to have got corrupted doing that.

dulciejackson (approved):

PR #318, opened 2021-01-28
GPG-629 Correct date on Action to close the gap page

AlexJacksonDS:
Noticed an incorrect date when checking over GPG-629 on dev


Binney (approved):
Noice

dulciejackson (approved):

PR #319, opened 2021-01-28
GPG-657 Specify a date format when creating csvs

AlexJacksonDS:
Specifies the date format as yyyy/MM/dd to output when creating csvs


dulciejackson (approved):
Niiiice 😃

Binney (approved):
Looks good to me!

PR #322, opened 2021-02-08
GPG-603 Some final tweaks to load testing setup and documentation

dulciejackson:
A few updates to the load testing README, the CSS selectors used in the load test code, and an extra check to make sure the Manage Organisation page loads as expected.

### DIFF
--- a/Infrastructure/CreateEnvironment.sh
+++ b/Infrastructure/CreateEnvironment.sh
@@ -14,7 +14,7 @@ REDIS_SIZE="tiny-4.x"
 # App scale settings
 APP_INSTANCES=2     # We use 2 for most environments, 3 for production
 APP_DISK="1G"       # Not sure what this default should be - we'll have to test this and set a default later
-APP_MEMORY="4G"     # In Azure this is 4G for all environments
+APP_MEMORY="4G"     # In PaaS this is 4G for all environments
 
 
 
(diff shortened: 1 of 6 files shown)


Infrastructure/CreateEnvironment.sh
```diff
@@ -14,7 +14,7 @@ REDIS_SIZE="tiny-4.x"
 # App scale settings
 APP_INSTANCES=2     # We use 2 for most environments, 3 for production
 APP_DISK="1G"       # Not sure what this default should be - we'll have to test this and set a default later
-APP_MEMORY="1G"     # In Azure this is 4G for most environments, 8G for production
+APP_MEMORY="4G"     # In Azure this is 4G for all environments
```
Binney, line 17: Since we don't use Azure we should probably change this to "PaaS"

LoadTests/src/test/scala/default/RecordingSimulation.scala
```diff
@@ -33,7 +33,7 @@ class RecordingSimulation extends Simulation {
 
 	val headers_3 = Map(
 		"Accept" -> "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8",
-		"Origin" -> "https://wa-t1pp-gpg.azurewebsites.net",
+		"Origin" -> "https://gender-pay-gap-loadtest.london.cloudapps.digital",
```
Binney, line 36: I'm a bit sad this is checked in to source control (and isn't e.g. environment config) but whatever

Binney (approved):
Works for me!

PR #323, opened 2021-02-08
GPG-603 Adds alternative reporting journey

dulciejackson:
Fixes some final stuff for the load test, including adding the alternative reporting journey so both versions of the reporting journey (swapped with a feature flag) can be tested.

### DIFF
--- a/Infrastructure/gpg-ipdeny/manifest.yml
+++ b/Infrastructure/gpg-ipdeny/manifest.yml
@@ -2,8 +2,8 @@
 
 applications:
   - name: gpg-ipdeny
-    instances: 3
-    memory: 3G
+    instances: 1
+    memory: 2G
 
     buildpacks:
       - nginx_buildpack
--- a/Infrastructure/gpg-ipdeny/manifest.yml
+++ b/Infrastructure/gpg-ipdeny/manifest.yml
@@ -2,8 +2,8 @@
 
 applications:
   - name: gpg-ipdeny
-    instances: 1
-    memory: 256M
+    instances: 3
+    memory: 3G
 
     buildpacks:
       - nginx_buildpack
--- a/Infrastructure/gpg-ipdeny/nginx.conf
+++ b/Infrastructure/gpg-ipdeny/nginx.conf
@@ -4,7 +4,7 @@ daemon off;
 error_log /dev/stderr;
 
 events {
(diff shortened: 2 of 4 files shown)


Infrastructure/gpg-ipdeny/manifest.yml
```diff
@@ -2,8 +2,8 @@
 
 applications:
   - name: gpg-ipdeny
-    instances: 1
-    memory: 256M
+    instances: 3
+    memory: 3G
```
Binney, line 6: Is this a valid number? I'd have expected it to be 2 or 4

Infrastructure/gpg-ipdeny/manifest.yml
```diff
@@ -2,8 +2,8 @@
 
 applications:
   - name: gpg-ipdeny
-    instances: 1
-    memory: 256M
+    instances: 3
```
Binney, line 5: Is this on all environments? feels very high for dev, test and preprod - would've expected those to be 1 each. And we're going to autoscale live anyway right?

Infrastructure/gpg-ipdeny/nginx.conf
```diff
@@ -4,7 +4,7 @@ daemon off;
 error_log /dev/stderr;
 
 events {
-  worker_connections 1024;
+  worker_connections 10000;
```
Binney, line 7: did this do anything in the end? if not I'd vote for keeping it as the default 1024

Binney (approved):

Infrastructure/gpg-ipdeny/manifest.yml
dulciejackson: Ah yeah, good point. We haven't deployed the IP deny app to test or preprod yet so I hadn't given them much thought - I'll take this back down to 1 for now!

Infrastructure/gpg-ipdeny/nginx.conf
dulciejackson: Yeah, it fixed my initial errors - now we're seeing timeouts and 500s but before I was seeing connections being dropped when it reached the 1024 threshold and couldn't take on more. 10000 was recommended on a StackOverflow post about configuring these values for live environments, which is why I went for it!

PR #324, opened 2021-02-08
GPG-646 filters bug

malinaslevoaca1:
**Issue:** 
When there were multiple filters of the same type added (eg "sector") the query wasn't built correctly. It was `param=value1,value2` instead of `param=value1&param=value2`

I fixed this by converting the existing query to a `NameValueCollection` - the values will be handled separately.

Also updated the docs as part of this PR

### DIFF
--- a/GenderPayGap.WebUI/Classes/Extensions/HtmlHelpers.cs
+++ b/GenderPayGap.WebUI/Classes/Extensions/HtmlHelpers.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Threading.Tasks;
+using System.Web;
 using GenderPayGap.Core.Classes.ErrorMessages;
 using GenderPayGap.Extensions;
 using GenderPayGap.WebUI.Classes.Attributes;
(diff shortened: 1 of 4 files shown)


GenderPayGap.WebUI/Classes/Extensions/HtmlHelpers.cs
```diff
@@ -62,9 +63,11 @@ public static async Task<IHtmlContent> PartialModelAsync<T>(this IHtmlHelper htm
         public static string WithQuery(this IUrlHelper helper, string actionName, object routeValues)
         {
             var newRoute = new NameValueCollection();
-            foreach (string key in helper.ActionContext.HttpContext.Request.Query.Keys)
+
+            if (helper.ActionContext.HttpContext.Request.QueryString.HasValue)
             {
-                newRoute[key] = helper.ActionContext.HttpContext.Request.Query[key];
+                var existingRoute = HttpUtility.ParseQueryString(helper.ActionContext.HttpContext.Request.QueryString.Value);
+                newRoute.Add(existingRoute);
```
Binney, line 70: Nice! Have you tested it with any other more complex query strings elsewhere?

Binney, line 70: What does it do if you have a weird query string like
```
?param1=value1&param1=value2,value3
```
?

Binney (approved):
All looks good to me! Got a quick question about testing but pending that very happy to merge in to master

GenderPayGap.WebUI/Classes/Extensions/HtmlHelpers.cs
malinaslevoaca1: I tested it locally, on the search results page. Any other ideas?

malinaslevoaca1: "value2,value3" is considered a single value - as expected.

The issue was with the query string builder.
We had "p=v1&p=v2" and when building `newRoute` we ended up with `newRoute[p]` = `v1,v2` (In theory it should be iterable (we iterate through that list of values - line 79) but for some reason it was seen as a single value)

Basically now we do the same thing (build a name value collection) using a different method.

Binney: Search results is the most complicated GET query I can think of in the site, so this works for me :+1:

PR #325, opened 2021-02-09
GPG-647 Remove registration given expired PIN

malinaslevoaca1:
See ticket [GPG-647](https://technologyprogramme.atlassian.net/browse/GPG-647)

- Since `PinSent` and `RequestPin` actions have been removed I added a new action `PinReset` where the organisation request is removed
- If the `PrivateManualRegistration` flag is disabled `PinSent` action is called. However, when PITP is enabled we should remove the action.

The other Custom Error Messages that have the action set to either `pin-sent` or `request-pin` aren't used anywhere.

### DIFF
--- a/GenderPayGap.WebUI/Controllers/OrganisationController.cs
+++ b/GenderPayGap.WebUI/Controllers/OrganisationController.cs
@@ -12,6 +12,7 @@ using GenderPayGap.Extensions.AspNetCore;
 using GenderPayGap.WebUI.BusinessLogic.Services;
 using GenderPayGap.WebUI.Classes;
 using GenderPayGap.WebUI.Classes.Services;
+using GenderPayGap.WebUI.ErrorHandling;
 using GenderPayGap.WebUI.Helpers;
 using GenderPayGap.WebUI.Models.Organisation;
 using GenderPayGap.WebUI.Repositories;
@@ -239,7 +240,7 @@ namespace GenderPayGap.WebUI.Controllers
         [HttpGet("~/activate-organisation/{id}")]
         public IActionResult ActivateOrganisation(string id)
         {
-            //Ensure user has completed the registration process
+            // Ensure user has completed the registration process
             IActionResult checkResult = CheckUserRegisteredOk(out User currentUser);
             if (checkResult != null)
             {
@@ -257,6 +258,16 @@ namespace GenderPayGap.WebUI.Controllers
             if (userOrg == null)
             {
                 return new HttpForbiddenResult($"User {currentUser?.EmailAddress} is not registered for employer id {organisationId}");
+            }
+
+            // TODO - Delete this once PITP is enabled
+            if (userOrg.HasExpiredPin())
+            {
+                userOrg.Organisation.UserOrganisations.Remove(userOrg);
+                DataRepository.Delete(userOrg);
+                DataRepository.SaveChanges();
+
+                throw new PinExpiredException();
             }
 
             // Ensure this organisation needs activation on the users account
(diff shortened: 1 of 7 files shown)


dulciejackson (changes requested):
I don't think we should be doing this by making changes to `BaseController` (one of the older files we're trying to stop using), nor changing the `CustomErrorMessages.config` file.

Rather than adding to the BaseController CheckUserRegisteredOk method, we should add an explicit check to the controller which directs to the PIN renewal - I believe that the journey starts in the `ActivateOrganisation` method of `OrganisationController`.

If the user has an expired PIN for the Organisation, we should use the new error system rather than editing the old error codes - we do this using the `CustomErrorPageExceptions.cs` file to define a custom exception and associated view, and then throwing the newly defined exception in the controller. It'll follow a similar pattern to the `PasswordResetCodeExpiredException` in that file - it's cleaner, more readable and we can design custom views for situations like this!

Happy to chat or pair if you'd like to go through this in more detail :)

malinaslevoaca1:
> I don't think we should be doing this by making changes to `BaseController` (one of the older files we're trying to stop using), nor changing the `CustomErrorMessages.config` file.
> 
> Rather than adding to the BaseController CheckUserRegisteredOk method, we should add an explicit check to the controller which directs to the PIN renewal - I believe that the journey starts in the `ActivateOrganisation` method of `OrganisationController`.
> 
> If the user has an expired PIN for the Organisation, we should use the new error system rather than editing the old error codes - we do this using the `CustomErrorPageExceptions.cs` file to define a custom exception and associated view, and then throwing the newly defined exception in the controller. It'll follow a similar pattern to the `PasswordResetCodeExpiredException` in that file - it's cleaner, more readable and we can design custom views for situations like this!
> 
> Happy to chat or pair if you'd like to go through this in more detail :)

I see your point. `ActivateOrganisation` uses the BaseController CheckUserRegisteredOk method. I can add the verification in the `OrganisationController` but this means we check the same thing twice

GenderPayGap.WebUI/Controllers/OrganisationController.cs
```diff
@@ -133,7 +134,7 @@ public IActionResult DeclareScope(DeclareScopeModel model, string id)
             if (userOrg.PINConfirmedDate == null)
             {
                 return new HttpForbiddenResult(
-                    $"User {currentUser?.EmailAddress} has not completed registeration for organisation {userOrg.Organisation.EmployerReference}");
+                    $"User {currentUser?.EmailAddress} has not completed registration for organisation {userOrg.Organisation.EmployerReference}");
```
Binney, line 137: :joy:

GenderPayGap.WebUI/Views/Errors/PinExpired.cshtml
```diff
@@ -0,0 +1,18 @@
...
+
+<div class="grid-row">
+    <div class="column-two-thirds">
+        <div class="govuk-box-highlight">
+            <h1 class="bold-large">Your PIN has expired</h1>
+        </div>
+        <div class="subsection panel-indent">
+            <div class="bold-small">Get a replacement PIN</div>
+        </div>
+        <p></p>
```
Binney, line 15: Is this an accident or does it change the styling?

Binney (approved):
I sat on this for nearly a month, sorry!! Looks good to me

GenderPayGap.WebUI/Views/Errors/PinExpired.cshtml
malinaslevoaca1: I copied the content from [Error.cshtml](https://github.com/cabinetoffice/gender-pay-gap/blob/main/GenderPayGap.WebUI/Views/Shared/Error.cshtml) as this was used for displaying the error.

I checked on Firefox, Chrome and smaller screens and I can't see a difference. I'll remove it

malinaslevoaca1:
@Binney can you please take a look at this PR? 😄

Binney (dismissed):
Ahhh I totally had 4 out of these 5 files read and hadn't signed it off!! My bad - good to go @malinaslevoaca1 👍

PR #326, opened 2021-02-10
GPG-669 Added autoscaling policy

malinaslevoaca1:
Since CPU Utilization isn't reliable I think we should use `throughput` as the autoscaling metric.
I had a look at the metrics in Grafana and 100 should be a good threshold (I _think_ we can also increase it)

**Note:** A script needs to be added to the deployment pipeline as well. I plan on using the same policy.

I haven't tested it yet.


Binney:
Nice! Is this just for the ipdeny component or will we be using the same policy json for the C# server too?

Infrastructure/gpg-ipdeny/autoscaling_policy.json
```diff
@@ -0,0 +1,22 @@
+﻿{
+  "instance_min_count": 1,
```
Binney, line 2: I would expect us to have different min_count and max_counts on each environment; is that feasible? how would we go about this?

Binney (approved):
Nice, looking forward to seeing the test :D

malinaslevoaca1:
> Nice! Is this just for the ipdeny component or will we be using the same policy json for the C# server too?

I'll use the same policy and we can update them later if needed

dulciejackson (approved):
Looks good! 
Like Sarah said we might want to have different criteria/max instance limits for the two apps, but we can look into that more once we've tested how this reacts during a load test 🙂

Infrastructure/gpg-ipdeny/autoscaling_policy.json
malinaslevoaca1: Good point! I updated the script to build the json policy

Infrastructure/gpg-ipdeny/deploy.sh
```diff
@@ -77,3 +91,10 @@ fi
 PROTECTED_APP_HOSTNAME="$(cf curl "v3/apps/$(cf app "${PROTECTED_APP_NAME}" --guid)/routes" | jq -r --arg APPS_DOMAIN "${APPS_DOMAIN}" '[.resources[] | select(.url | endswith($APPS_DOMAIN))][0].host')"
 
 cf bind-route-service "${APPS_DOMAIN}" "${ROUTE_SERVICE_NAME}" --hostname "${PROTECTED_APP_HOSTNAME}";
+
+# Autoscaling
+echo "{\"instance_min_count\":${MIN_COUNT_INSTANCES},\"instance_max_count\":${MAX_COUNT_INSTANCES},\"scaling_rules\":[{\"metric_type\":\"throughput\",\"breach_duration_secs\":60,\"threshold\":100,\"operator\":\"<\",\"cool_down_secs\":60,\"adjustment\":\"-1\"},{\"metric_type\":\"throughput\",\"breach_duration_secs\":60,\"threshold\":100,\"operator\":\">=\",\"cool_down_secs\":60,\"adjustment\":\"+1\"}]}" > autoscaling_policy.json
```
Binney, line 96: Is there a way of doing this which doesn't result in overwriting the file that's already in source control? e.g. putting some placeholders in the JSON and then here in the bash script you just read the file into a variable and replace the placeholders

Binney (approved):

Infrastructure/gpg-ipdeny/deploy.sh
malinaslevoaca1: I removed the file but I agree - that's a better idea 👍

PR #329, opened 2021-02-12
GPG-NONE exit block of actions on fail

malinaslevoaca1:
If a request fails the others shouldn't be sent cause there is a dependency between them (e.g. "organisationId" may not be set)


dulciejackson (approved):
Awesome, thanks for handling this! 😄

PR #330, opened 2021-02-12
GPG-682 Update load test README

dulciejackson:
Added a bunch more context to the load test README from experience setting up infrastructure, db and configuring code.


Binney (approved):
Phew!! Great stuff :+1:

PR #331, opened 2021-02-19
Revert "GPG-164 Admin page to re-add retired submissions (will only b…

AlexJacksonDS:
…e used once, then removed)"

This reverts commit 60635aa40c07540c2990a3232054fe02b8e8a420.


Binney:
Can I ask you to add a quick comment around any testing you've done here? Also, as we don't tend to squash commits, is this the only commit we need to revert?

AlexJacksonDS:
Yeah this is the only commit that needs reverting to remove the functionality, it was all added in a single commit.

For testing, I just confirmed that the endpoint is no longer accessible, and a bit of clicking around the rest of the site. It wasn't linked anywhere so we have no broken links.

Binney (approved):
Awesome stuff, thanks

dulciejackson (approved):
Good stuff! :)

PR #332, opened 2021-02-22
Gpg 631 create account content changes

AlexJacksonDS:
Content changes and the new back/cancel buttons/links on the account creation screens.

I've gone through the site locally and made sure that the back/cancel links direct to the correct pages as expected when you are within the journey that starts at /already-created-an-account-question (linked from the Gov UK pages) or when you have accessed the create account page from an alternate route, e.g. the login page.

In the journey: 
- /already-created-an-account-question cancel links to the Gov UK StartUrl page (that links to it)
- /create-user-account back link takes you back to the /already-created-an-account-question page
- /create-user-account cancel links back to the Gov UK StartUrl page

Outside the journey:
- /create-user-account cancel links back to the ManageOrganisations endpoint (which if you aren't signed in takes you to the login page) and the back link doesn't display

### DIFF
--- a/GenderPayGap.WebUI/Controllers/Account/AccountCreationController.cs
+++ b/GenderPayGap.WebUI/Controllers/Account/AccountCreationController.cs
@@ -44,7 +44,7 @@ namespace GenderPayGap.WebUI.Controllers.Account
 
                 case HaveYouAlreadyCreatedYourUserAccount.No:
                 case HaveYouAlreadyCreatedYourUserAccount.NotSure:
-                    return RedirectToAction("CreateUserAccountGet", new { isPartOfGovUkReportingJourney = true});
+                    return RedirectToAction("CreateUserAccountGet", new { isPartOfGovUkReportingJourney = true });
 
                 case HaveYouAlreadyCreatedYourUserAccount.Unspecified:
                     viewModel.AddErrorFor(
@@ -58,7 +58,7 @@ namespace GenderPayGap.WebUI.Controllers.Account
                     return View("AlreadyCreatedAnAccountQuestion", model);
             }
         }
-        
+
         [HttpGet("/create-user-account")]
         public IActionResult CreateUserAccountGet(bool isPartOfGovUkReportingJourney = false)
         {
@@ -67,7 +67,7 @@ namespace GenderPayGap.WebUI.Controllers.Account
                 return RedirectToAction("ManageOrganisationsGet", "ManageOrganisations");
             }
 
-            return View("CreateUserAccount", new CreateUserAccountViewModel{IsPartOfGovUkReportingJourney = isPartOfGovUkReportingJourney });
+            return View("CreateUserAccount", new CreateUserAccountViewModel { IsPartOfGovUkReportingJourney = isPartOfGovUkReportingJourney });
         }
 
         [HttpPost("/create-user-account")]
@@ -114,7 +114,7 @@ namespace GenderPayGap.WebUI.Controllers.Account
                 {
                     viewModel.AddErrorFor(
                         m => m.EmailAddress,
-                        "This email address is awaiting confirmation. Please check you email inbox or enter a different email"
+                        "This email address is awaiting confirmation. Please check your email inbox or enter a different email"
                         + " address");
                 }
             }
@@ -130,20 +130,20 @@ namespace GenderPayGap.WebUI.Controllers.Account
 
             GenerateAndSendAccountVerificationEmail(newUser);
 
-            var confirmEmailAddressViewModel = new ConfirmEmailAddressViewModel {EmailAddress = viewModel.EmailAddress};
+            var confirmEmailAddressViewModel = new ConfirmEmailAddressViewModel { EmailAddress = viewModel.EmailAddress };
             return View("ConfirmEmailAddress", confirmEmailAddressViewModel);
         }
-        
+
         [HttpGet("/verify-email")]
         public IActionResult VerifyEmail(string code)
         {
             User gpgUser = dataRepository.GetAll<User>().FirstOrDefault(u => u.EmailVerifyHash == code);
-            
+
             if (gpgUser == null)
             {
                 return View("UserNotFoundErrorPage");
             }
-            
+
             if (User.Identity.IsAuthenticated || gpgUser.EmailVerifiedDate != null)
             {
                 return RedirectToAction("ManageOrganisationsGet", "ManageOrganisations");
@@ -152,7 +152,7 @@ namespace GenderPayGap.WebUI.Controllers.Account
             gpgUser.EmailVerifiedDate = VirtualDateTime.Now;
             gpgUser.SetStatus(UserStatuses.Active, gpgUser, "Email verified");
             dataRepository.SaveChanges();
-            
+
             return RedirectToAction("AccountCreationConfirmation");
         }
 
@@ -203,7 +203,7 @@ namespace GenderPayGap.WebUI.Controllers.Account
             string verificationUrl = Url.Action(
                 "VerifyEmail",
                 "AccountCreation",
-                new {code = verificationCode},
+                new { code = verificationCode },
                 "https");
 
             try
--- a/GenderPayGap.WebUI/Controllers/Account/AccountCreationController.cs
+++ b/GenderPayGap.WebUI/Controllers/Account/AccountCreationController.cs
@@ -26,7 +26,9 @@ namespace GenderPayGap.WebUI.Controllers.Account
             this.dataRepository = dataRepository;
             this.emailSendingService = emailSendingService;
         }
-        
+
+        // The 'Start Now' page (Global.StartUrl https://www.gov.uk/report-gender-pay-gap-data)
+        // links to this action as the starting point for the reporting journey
         [HttpGet("/already-created-an-account-question")]
         public IActionResult AlreadyCreatedAnAccountQuestionGet(AlreadyCreatedAnAccountViewModel viewModel)
         {
@@ -42,7 +44,7 @@ namespace GenderPayGap.WebUI.Controllers.Account
 
                 case HaveYouAlreadyCreatedYourUserAccount.No:
                 case HaveYouAlreadyCreatedYourUserAccount.NotSure:
-                    return RedirectToAction("CreateUserAccountGet");
+                    return RedirectToAction("CreateUserAccountGet", new { isPartOfGovUkReportingJourney = true});
 
                 case HaveYouAlreadyCreatedYourUserAccount.Unspecified:
                     viewModel.AddErrorFor(
@@ -58,14 +60,14 @@ namespace GenderPayGap.WebUI.Controllers.Account
         }
         
         [HttpGet("/create-user-account")]
-        public IActionResult CreateUserAccountGet()
+        public IActionResult CreateUserAccountGet(bool isPartOfGovUkReportingJourney = false)
         {
             if (User.Identity.IsAuthenticated)
             {
                 return RedirectToAction("ManageOrganisationsGet", "ManageOrganisations");
             }
 
-            return View("CreateUserAccount", new CreateUserAccountViewModel());
+            return View("CreateUserAccount", new CreateUserAccountViewModel{IsPartOfGovUkReportingJourney = isPartOfGovUkReportingJourney });
         }
 
         [HttpPost("/create-user-account")]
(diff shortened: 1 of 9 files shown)


GenderPayGap.WebUI/Views/AccountCreation/AlreadyCreatedAnAccountQuestion.cshtml
```diff
@@ -76,15 +76,18 @@
...
-                    organisation's gender pay gap.
+                    You will need to create your own user account even if someone else has
+                    previously reported your employer’s gender pay gap.
                 </div>
 
                 @(Html.GovUkButton(new ButtonViewModel
                 {
                     Text = "Continue"
                 }))
+                @{ var reportingUrl = Url.Action("ManageOrganisationsGet", "ManageOrganisations"); }
```
malinaslevoaca1, line 87: Why do we want the user to be redirected to the manage organisations page if he doesn't have an account?

GenderPayGap.WebUI/Controllers/Account/AccountCreationController.cs
```diff
@@ -58,14 +60,14 @@ public IActionResult AlreadyCreatedAnAccountQuestionGet(AlreadyCreatedAnAccountV
...
-        public IActionResult CreateUserAccountGet()
+        public IActionResult CreateUserAccountGet(bool isPartOfGovUkReportingJourney = false)
         {
             if (User.Identity.IsAuthenticated)
             {
                 return RedirectToAction("ManageOrganisationsGet", "ManageOrganisations");
             }
 
-            return View("CreateUserAccount", new CreateUserAccountViewModel());
+            return View("CreateUserAccount", new CreateUserAccountViewModel{IsPartOfGovUkReportingJourney = isPartOfGovUkReportingJourney });
```
malinaslevoaca1, line 70: Formatting is a bit off (missing spaces before and after `{`)

GenderPayGap.WebUI/Models/AccountCreation/AlreadyCreatedAnAccountViewModel.cs
```diff
@@ -20,7 +20,7 @@ public enum HaveYouAlreadyCreatedYourUserAccount
         [GovUkRadioCheckboxLabelText(Text = "No, I don't have a user account")]
         No = 1,
 
-        [GovUkRadioCheckboxLabelText(Text = "My organisation has used this service before, but I haven't")]
+        [GovUkRadioCheckboxLabelText(Text = "My employer has used this service before, but I haven't")]
```
malinaslevoaca1, line 23: There's an inconsistency in the apostrophes (`’` vs `'`)

malinaslevoaca1 (commented):
There was an error message change requested. Shouldn't that be included as well?

GenderPayGap.WebUI/Views/AccountCreation/AlreadyCreatedAnAccountQuestion.cshtml
AlexJacksonDS: I've just copied the same behaviour as the rest of the places where a similar pattern is used. 

If they don't have an account/aren't signed in it'll just redirect to the login page, if they are signed in and ended up here somehow, then that Manage Organisations page is the same place that login takes you.

For live, this isn't really relevant anyway since Global.StartUrl isn't null, we'll only see this behaviour on the other environments

GenderPayGap.WebUI/Models/AccountCreation/AlreadyCreatedAnAccountViewModel.cs
AlexJacksonDS: I think that's just a consequence of the font used when drafting the changes, we use ' everywhere else on the site, the only word that has changed here is organisation to employer.

AlexJacksonDS:
> There was an error message change requested. Shouldn't that be included as well?
Yeah sure I can do them here instead while I'm making other changes. There is also GPG-654 that covers changing them for all these content tickets.

GenderPayGap.WebUI/Controllers/Account/AccountCreationController.cs
```diff
@@ -128,20 +130,20 @@ public IActionResult CreateUserAccountPost(CreateUserAccountViewModel viewModel)
...
+            var confirmEmailAddressViewModel = new ConfirmEmailAddressViewModel { EmailAddress = viewModel.EmailAddress };
             return View("ConfirmEmailAddress", confirmEmailAddressViewModel);
         }
-        
+
         [HttpGet("/verify-email")]
         public IActionResult VerifyEmail(string code)
         {
             User gpgUser = dataRepository.GetAll<User>().FirstOrDefault(u => u.EmailVerifyHash == code);
-
```
Binney, line 139: oh my god thank you this is so satisfying

GenderPayGap.WebUI/Models/AccountCreation/AlreadyCreatedAnAccountViewModel.cs
Binney: FWIW I think it should be `’` (right single quotation mark) in guidance text ([citation](https://www.cl.cam.ac.uk/~mgk25/ucs/quotes.html)) and the apostrophes `'` are just because that's easier to type. Agree consistency is more important than unicode precision though...!

Binney (approved):
Nice!

GenderPayGap.WebUI/Views/AccountCreation/AlreadyCreatedAnAccountQuestion.cshtml
dulciejackson: Exactly - the ManageOrganisations page checks whether the user is signed in, and if not they're auto-directed back to the sign in page. This actually happens in a number of places on the service (and we've got helper methods for it), but ManageOrganisations generally feels like a sensible 'home page' style place for logged in users, which it a good place to check if the user needs to sign in.

GenderPayGap.WebUI/Views/AccountCreation/AlreadyCreatedAnAccountQuestion.cshtml
```diff
@@ -76,15 +76,18 @@
...
+                <a href="@(Global.StartUrl ?? reportingUrl)" class="govuk-button" role="button">Cancel</a>
```
dulciejackson, line 88: We use links for 'Cancel' buttons, but we always style them as links rather than buttons. 
The pattern can be seen in `ChangeContactPreferences.cshtml` on lines 68-72. We need to surround links with `p` body styling so the text is formatted correctly - particularly so the font size is inherited.

GenderPayGap.WebUI/Views/AccountCreation/ConfirmationPage.cshtml
```diff
@@ -26,23 +26,27 @@
...
+            You are now able to add your employer to your user account and
+            report their gender pay gap information. 
         </p>
         
-        <h2 class="govuk-heading-m">Editing your contact details or closing your account</h2>
+        <h2 class="govuk-heading-m">Editing your contact details or closing your user account</h2>
         
         <p class="govuk-body">
-            You can edit your contact details within Manage My Account.
+            You can edit your contact details within Manage Account.
```
dulciejackson, line 36: I know Claire didn't mention it, but we could link the Manage Account text here to the actual page? We use that pattern elsewhere on the service and finding the Manage Account page isn't entirely clear (it's the second tab on Manage Organisations), so it'd probably help users to navigate the service from here 😄

GenderPayGap.WebUI/Views/AccountCreation/ConfirmationPage.cshtml
```diff
@@ -26,23 +26,27 @@
...
+                <li>created a user account by mistake</li>
+                <li>
+                    you no longer work for, or manage the gender pay gap reporting for
+                    the employer added to your account
+                </li>
             </ul>
         </p>
         <p class="govuk-body">
-            To complete the process for gender pay gap reporting please continue.
+            To complete the process for gender pay gap reporting please sign in to your account.
```
dulciejackson, line 49: Will the user not already be signed in to their account at this point?

Again, we could link this 'sign in to your account' text to the sign in page.

GenderPayGap.WebUI/Views/AccountCreation/CreateUserAccount.cshtml
```diff
@@ -149,29 +156,28 @@
...
-
                 @(Html.GovUkButton(new ButtonViewModel
                 {
-                    Text = "Continue",
+                    Text = "Save and Continue",
                     Classes = "govuk-!-margin-top-9"
                 }))
-
+                @{ var reportingUrl = Url.Action("ManageOrganisationsGet", "ManageOrganisations"); }
+                <a href="@(Model.IsPartOfGovUkReportingJourney ? Global.StartUrl ?? reportingUrl : reportingUrl)" class="govuk-button govuk-!-margin-top-9" role="button">Cancel</a>
```
dulciejackson, line 167: Again, this should be styled as a link, not a button, with the `govuk-link` class

dulciejackson (changes requested):
Looking good! Just a few thoughts 😄

GenderPayGap.WebUI/Views/AccountCreation/ConfirmationPage.cshtml
AlexJacksonDS: No it doesn't log in when verifying the email, it just has a button link to the sign in page below it. Not sure if it is worth adding a second link?

GenderPayGap.WebUI/Views/AccountCreation/ConfirmationPage.cshtml
AlexJacksonDS: Adding this link would just redirect to the login page since the email confirmation doesn't log you in, probably not worth adding unless we want to change that behaviour

Binney (approved):

GenderPayGap.WebUI/Views/AccountCreation/AlreadyCreatedAnAccountQuestion.cshtml
```diff
@@ -85,7 +85,9 @@
                     Text = "Continue"
                 }))
                 @{ var reportingUrl = Url.Action("ManageOrganisationsGet", "ManageOrganisations"); }
-                <a href="@(Global.StartUrl ?? reportingUrl)" class="govuk-button" role="button">Cancel</a>
+                <p class="govuk-body">
+                    <a href="@(Global.StartUrl ?? reportingUrl)" class="govuk-link" role="button">Cancel</a>
```
dulciejackson, line 89: We don't need the `role="button"` here since a screen-reader would be right to read it out as a link!

GenderPayGap.WebUI/Views/AccountCreation/CreateUserAccount.cshtml
```diff
@@ -164,7 +164,9 @@
                     Classes = "govuk-!-margin-top-9"
                 }))
                 @{ var reportingUrl = Url.Action("ManageOrganisationsGet", "ManageOrganisations"); }
-                <a href="@(Model.IsPartOfGovUkReportingJourney ? Global.StartUrl ?? reportingUrl : reportingUrl)" class="govuk-button govuk-!-margin-top-9" role="button">Cancel</a>
+                <p class="govuk-body">
+                    <a href="@(Model.IsPartOfGovUkReportingJourney ? Global.StartUrl ?? reportingUrl : reportingUrl)" class="govuk-link govuk-!-margin-top-9" role="button">Cancel</a>
```
dulciejackson, line 168: Again, we don't need the `role="button"` here 😄

GenderPayGap.WebUI/Views/AccountCreation/AlreadyCreatedAnAccountQuestion.cshtml
AlexJacksonDS: ah whoops i'll fix them

GenderPayGap.WebUI/Views/AccountCreation/ConfirmationPage.cshtml
dulciejackson: Ah yeah ok cool, makes sense!

GenderPayGap.WebUI/Views/AccountCreation/ConfirmationPage.cshtml
dulciejackson: Ok cool - in that case I think the button is fine then (and we don't need to link here as well)

dulciejackson (approved):

PR #333, opened 2021-02-23
GPG-632 Content and route changes for Add Employer

AlexJacksonDS:
Content changes for the Add Employer journey. As part of this work I've also changed the routes from /add-organisation/ to /add-employer/ for all of these pages. This means that we have a new controller containing all the redirects for the old urls, in case they were bookmarked or linked from somewhere external for example (unlikely since it is all in the account protected area).

I've gone through the journey a few times and it all seems fine, and I didn't find anywhere that was linking to these pages with a none generated link, so everything should be fine.


Binney (approved):
Noiceeeeeeee

dulciejackson (commented):
Do we want to swap some of these file names to use the new 'Employer' terminology rather than 'Organisation'? Keen to keep up-to-date but also don't want to cause language conflicts here (with either contents of files _or_ commonly used properties/db tables) - happy to discuss this one, I've got no strong opinions either way 😄

AlexJacksonDS:
New ticket to handle the renaming [GPG-737](https://technologyprogramme.atlassian.net/browse/GPG-737) merging this one

PR #334, opened 2021-02-23
Gpg 650 manage account content changes

AlexJacksonDS:
Content changes for the Manage Account pages. There were also a few areas that weren't included in the pdf document that I have updated in line with those that were (Organisation to Employer)

Only a few changes here, all wording. I've checked that they appear as expected. Note that the change for the Manage Organisation tab to Manage Employer has been done in GPG-632.


Binney (approved):
I love it when code is actually simple

dulciejackson (approved):

PR #335, opened 2021-02-25
Gpg 677 increase live memory allowance

AlexJacksonDS:
The plan is to update the 'Push app to Gov.UK PaaS' step in the Deploy to PaaS (push) task group to 
`cf push $(AppName) --strategy rolling -f manifest-$(SpaceName).yml`

The SpaceName variable already exists in Azure (set to gpg-dev etc.), so I'm using those and naming the manifest files accordingly rather than defining a new set of variables.

I've just added the instances/memory as they are currently set in Gov PaaS. Not really sure how best we can test this as it'll mean that we are changing all the deployment pipelines due to common steps. i.e. we may not be able to deploy older stuff to prod without reverting the change in Azure.


malinaslevoaca1 (commented):
I think we want to have the same config for all the other environments, right? If so, can't we have 2 `manifest` files - one for Prod and one for the others? (Ignore the load test environment - it can use the other config as well, I made some changes manually to replicate the live env)

AlexJacksonDS:
> I think we want to have the same config for all the other environments, right? If so, can't we have 2 `manifest` files - one for Prod and one for the others? (Ignore the load test environment - it can use the other config as well, I made some changes manually to replicate the live env)

I think it'll depend on how much we want to configure the different environments differently. Would have to change the approach slightly with the cf push command and pipeline variables

Binney (approved):
Looks good to me! I sort of agree with Malina but I actually also think for now we should keep them separate and set Preprod to 8GB too, so we can practice the zero downtime release (not like I think this will actually be an issue).

AlexJacksonDS:
Just to make sure it's clear what is happening here:

- Updated the preprod memory to 8GB
- All environments are separate manifest files, if we want to change this to 2 files in the future as proposed by Malina, there will have to be a further pipeline change (a small one)
- This needs merging before I can save any pipelines changes otherwise it'll try to find files that aren't there

@malinaslevoaca1 @Binney Are we happy with that?

PR #336, opened 2021-03-03
GPG-709 Changed deadline for 2020 reporting year

malinaslevoaca1:
Ticket [GPG-709](https://technologyprogramme.atlassian.net/browse/GPG-709)

- Add 6 months to the deadline if accounting year is 2020
- Moved logic to a helper class
- Updated `SendReminderEmails` job to use the helper method as well - this means we will be able to use the exact same template without making further changes (needs to be tested, though)

Tested locally:
- employer page (public and private)
- manage organisation (public & private)
- late submission
- Admin CSV Files (late submissions, full submission history, all submissions)

To be tested:
- reminder jobs (not a priority - [GPG-711](https://technologyprogramme.atlassian.net/browse/GPG-711) can be converted to a test ticket)
- public facing CSV files - these files are built by a background job - I'll check this once deployed to dev


GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
```diff
@@ -116,10 +117,11 @@
                                         {
                                             if (organisation.GetIsInscope(report.AccountingDate.Year))
                                             {
+                                                DateTime deadline = ReportingYearsHelper.GetDeadlineForAccountingDate(report.AccountingDate);
                                                 <span>
                                                     This organisation
```
AlexJacksonDS, line 122: It's probably covered by the other content tickets (the 6 I'm halfway through) but we should be changing organisation to employer. This is perhaps not the best ticket to do it on though, might as well group it with the rest of those changes

AlexJacksonDS (approved):

GenderPayGap.WebUI/Helpers/ReportingYearsHelper.cs
```diff
@@ -29,5 +30,18 @@ public static string FormatYearAsReportingPeriod(int reportingPeriodStartYear)
             return formattedYear;
         }
 
+        public static DateTime GetDeadlineForAccountingDate(DateTime accountingDate)
+        {
+            const int deadlineExtensionFor2020InMonths = 6;
```
Binney, line 35: I'd have expected this to go at the top of the file and maybe be in all caps (but maybe I've spent too much time staring at Java)

More helpfully, I definitely would've expected it to be commonised between here and the Return model class

GenderPayGap.WebUI/Helpers/ReportingYearsHelper.cs
```diff
@@ -29,5 +30,18 @@ public static string FormatYearAsReportingPeriod(int reportingPeriodStartYear)
             return formattedYear;
         }
 
+        public static DateTime GetDeadlineForAccountingDate(DateTime accountingDate)
```
Binney, line 33: BIG FAN

Binney (approved):
Big yes

Binney (approved):
Perfecto, like it

PR #337, opened 2021-03-03
GPG-680 Update IP Deny README with automated deployment details

AlexJacksonDS:
Updated the readme for the automated deployments of the IP deny app.

I've made changes to:

- the build all branches job to include and IP Deny Artifact
- the Deploy to PaaS (prepare) task group to include getting this new aritifact
- the Deploy to PaaS (push) task group to get the new secret file and deploy the IP Deny app
- Added a bunch of variables to the Deploy to PaaS/DEV release pipeline

Both the task groups are version 2.\*, so when we are ready to promote these to other environments, we should just need to add the new variables to the relevant releases and update the 2 task groups from version 1.\* to 2.*

I've also attempted to fix the automated deployment to dev when main is updated, which was broken by the change from main to master, but that may not have worked (the manually triggered deployments still work though if that hasn't worked).

I've tested that the deployments work, and deploy the IP deny app successfully given that the dev IP Deny app was not the same as the one that was subsequently deployed to loadtest (no autoscaling), so it should work, but I cant be sure until we deploy it to to an environment that doesn't yet have it (e.g. preprod)

(edited to fix accidental styling)

### DIFF
--- a/GenderPayGap.WebUI/GenderPayGap.WebUI.csproj
+++ b/GenderPayGap.WebUI/GenderPayGap.WebUI.csproj
@@ -108,6 +108,9 @@
     <ProjectReference Include="..\GovUkDesignSystem\GovUkDesignSystem.csproj" />
   </ItemGroup>
   <ItemGroup>
+    <None Update="AzureAttachAutoscalingPolicy.sh">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
     <None Update="buildpack.yml">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
--- a/Infrastructure/gpg-ipdeny/README.md
+++ b/Infrastructure/gpg-ipdeny/README.md
@@ -4,7 +4,7 @@ gpg-ipdeny is an IP denylisting nginx application based upon the IP allowing ver
 
 It is designed to allow access to the Gender Pay Gap Service to all IP addresses except those specifically set, returning a 403 response to those banned IPs.
 
-## Deployment
+## Manual Deployment
 
 Since the list of banned IPs should not update particularly frequently, the application is deployed manually via the Deploy.sh script.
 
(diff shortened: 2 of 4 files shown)


malinaslevoaca1 (dismissed):
Great work! 👍

Binney:
Fab stuff!

Given we have a tragically small amount of the Azure DevOps pipeline actually in source control, how would you feel about moving `Deploy the IP Deny application` into a bash script in code? (It looks like you can then use it with a "File path" instead of "inline") And also while you're at it doing the "Attach autoscaling policy" one?

Infrastructure/gpg-ipdeny/README.md
```diff
@@ -28,10 +28,18 @@ The deployment script takes the following parameters:
...
 
 When updating this list, each IP should be listed on a new line.
-If there are any errors when starting the app, these are most likely to have been caused by a syntax error in the list of IPs to deny. By checking the environment variables that have been set on the environment, it should be possible to tell where the issue is in the provided list of IPs
\ No newline at end of file
+If there are any errors when starting the app, these are most likely to have been caused by a syntax error in the list of IPs to deny. By checking the environment variables that have been set on the environment, it should be possible to tell where the issue is in the provided list of IPs.
+There is also a copy of the GPG-IP-Denylist.txt file stored in the Library - Secure Files of Azure. If the list of IPs is updated, this should be updated along with the version in Zoho.
+
+## Automated deployment
+The IP deny app is deployed via the same Azure release pipelines using a Bash step in the Deploy to PaaS (push) task group
```
Binney, line 41: Niceeeeeeee

Binney (dismissed):

GenderPayGap.WebUI/GenderPayGap.WebUI.csproj
```diff
@@ -108,6 +108,9 @@
     <ProjectReference Include="..\GovUkDesignSystem\GovUkDesignSystem.csproj" />
   </ItemGroup>
   <ItemGroup>
+    <None Update="AzureAttachAutoscalingPolicy.sh">
```
malinaslevoaca1, line 111: Don't we need to include `AzureDeployIPDeny.sh` as well?

AlexJacksonDS: It's not included in the WebUI, there's a separate export step that handles the IP deny stuff that will get it

malinaslevoaca1: I see, thanks!

malinaslevoaca1 (approved):

PR #339, opened 2021-03-04
GPG=676 Updated threshold

malinaslevoaca1:
This is the threshold used on the load test environment.

A new instance is added if throughput is greater than 10rps for 60 seconds

As part of this ticket I also added a task to the[ Deploy to PaaS (push) group](https://dev.azure.com/govtequalitiesoffice/Gender%20Pay%20Gap/_taskgroup/b7fe281e-469b-4452-9be0-58a5f1db2e04) for attaching the policy to the main app.

### DIFF
--- a/Infrastructure/gpg-ipdeny/autoscaling_policy_template.json
+++ b/Infrastructure/gpg-ipdeny/autoscaling_policy_template.json
@@ -5,7 +5,7 @@
     {
       "metric_type": "throughput",
       "breach_duration_secs": 60,
-      "threshold": 100,
+      "threshold": 10,
       "operator": "<",
       "cool_down_secs": 60,
       "adjustment": "-1"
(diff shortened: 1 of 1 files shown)


AlexJacksonDS (commented):
Looks like you are using the same Min/Max instances count as the IP Deny app for the main app. Is that the intended values?

malinaslevoaca1:
> Looks like you are using the same Min/Max instances count as the IP Deny app for the main app. Is that the intended values?

We used the same values on the load test environment. I'd rather have the same values at least for now (i.e. have multiple IP deny instances in Prod) unless you think otherwise.

AlexJacksonDS:
> We used the same values on the load test environment. I'd rather have the same values at least for now (i.e. have multiple IP deny instances in Prod) unless you think otherwise.

Yeah multiple IP deny instances in prod is fine, and is handled by the stuff I did. This will set the number of instances of the web app to the same values, unless I've misunderstood what the azure step is doing

Binney (approved):

PR #341, opened 2021-03-05
GPG-NONE Fixed errors

malinaslevoaca1:
We got some errors after rebase due to missing imports


AlexJacksonDS (approved):
Just to check if it was something I overlooked, was this caused by one of the merges I just did? I changed those files recently

malinaslevoaca1:
@AlexJacksonDS You added a variable imported from `GenderPayGap.Core` - I removed this import cause the existing stuff has been moved to `Core.Helpers` so during rebase the import was removed

PR #346, opened 2021-03-10
Gpg 732 update reporting deadline 2020

AlexJacksonDS:
Changed the 2020 deadline date to 05/10/2021 (for both public and private)

Also updated the draft return deadline checking so that it checks whether it is late using the new deadline helper method. This wouldn't affect 2020 as it has been added to the late flag list so the overall method will still evaluate to false. I've also added some comments around the confusing deadline vs due date naming we have so that it should be clearer why we are doing things like adding days

Checked that it displays 5th October as expected for public/private and that the logic will work consistently for deciding if a submission is late.

(I had some trouble with git not adding all the files I wanted to the commit, so I've ended up with some weird ones)


malinaslevoaca1 (approved):
Looks good! Thanks for spotting the `DraftReturnService` use of the deadline.

PR #348, opened 2021-03-12
GPG-734 Add scopes for all the reporting years when adding a new orga…

malinaslevoaca1:
https://technologyprogramme.atlassian.net/browse/GPG-734


Binney (approved):
LGTM, will need to revisit when we revamp the reporting years system in May but can wait until then!

PR #349, opened 2021-03-16
GPG-728 Fixed decimals rounded in edit mode issue

malinaslevoaca1:
Ticket [GPG-728](https://technologyprogramme.atlassian.net/browse/GPG-728)

The decimals were truncated to 1 decimal place because we applied the format string in both display and edit mode.
Even if we removed this, the decimals would be truncated to 2 decimal places in edit mode because this is the [default behaviour](https://github.com/aspnet/AspNetWebStack/blob/42991b3d2537b702736463f76a10a4fcf2ea44c9/src/System.Web.Mvc/Html/DefaultEditorTemplates.cs#L110).

As a workaround, we can use the "String" editor template. By doing this, on error, the user input is displayed and not a rounded value. User is still not able to submit the form if a figure has more than 1 decimal place. When displayed (on the review report page) the figures have a decimal place as before.


Binney (approved):
LGTM, does this increase the risk we might end up with bad data in those fields? i.e. someone types words into them

malinaslevoaca1:
> LGTM, does this increase the risk we might end up with bad data in those fields? i.e. someone types words into them

No, it's the same behavior as before. User can type words but those values are ignored

PR #350, opened 2021-03-16
GPG-716 Allow a user to add a retired employer

AlexJacksonDS:
Agreed that we should always just be re-activating the employer where it exists since the case where there are deleted/retired employers with submissions is only for retired employers which appear in the search to be able to select anyway.

There will be no issues with displaying the incorrect data in this case, as either the employer will have no submission data, or it will be one of these searchable retired employers perhaps with some data that would still be relevant since it is the same employer. Agreed that the process for manual deletion should also mark the submissions as retired where necessary, so we should never be in the case where incorrect data will appear.

Tested adding a few employers in various statuses to my local account and it either works as before, or successfully re-activates/logs as expected rather than throwing the duplicate CompanyNumber entry exception we currently see on live.

### DIFF
--- a/GenderPayGap.WebUI/Services/OrganisationService.cs
+++ b/GenderPayGap.WebUI/Services/OrganisationService.cs
@@ -20,16 +20,18 @@ namespace GenderPayGap.WebUI.Services
             Manual
         }
 
-
+        private readonly AuditLogger auditLogger;
         private readonly IDataRepository dataRepository;
         private readonly ICompaniesHouseAPI companiesHouseApi;
 
         public OrganisationService(
             IDataRepository dataRepository,
-            ICompaniesHouseAPI companiesHouseApi)
+            ICompaniesHouseAPI companiesHouseApi,
+            AuditLogger auditLogger)
         {
             this.dataRepository = dataRepository;
             this.companiesHouseApi = companiesHouseApi;
+            this.auditLogger = auditLogger;
         }
 
         public void UpdateIsUkAddressIfItIsNotAlreadySet(long organisationId, bool? isUkAddress)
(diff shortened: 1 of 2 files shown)


GenderPayGap.WebUI/Services/OrganisationService.cs
```diff
@@ -134,6 +125,49 @@ public Organisation ImportOrganisationFromCompaniesHouse(string companyNumber, U
             return organisation;
         }
 
+        private Organisation SetDeletedRetiredOrganisationToActive(Organisation existingOrganisation, User requestingUser)
+        {
+            var oldStatus = existingOrganisation.Status;
+
+            existingOrganisation.SetStatus(OrganisationStatuses.Active, requestingUser.UserId, "Automatic reactivation");
```
malinaslevoaca1, line 132: What if the company data (e.g., the address) has been updated in the meantime? Should we do that as well?

GenderPayGap.WebUI/Services/OrganisationService.cs
```diff
@@ -134,6 +125,49 @@ public Organisation ImportOrganisationFromCompaniesHouse(string companyNumber, U
             return organisation;
         }
 
+        private Organisation SetDeletedRetiredOrganisationToActive(Organisation existingOrganisation, User requestingUser)
```
malinaslevoaca1, line 128: Let's say an organisation had been deleted/retired before the latest scopes were added. Does the organisation have all the scopes?

I assume no scopes are added unless an organisation is active. If so, we need to add any missing scopes here as well.

malinaslevoaca1 (commented):
Nice work! A couple of questions:

GenderPayGap.WebUI/Services/OrganisationService.cs
AlexJacksonDS: There is a job FetchCompaniesHouseDataJob that handles this kind of thing (not sure on the schedule) that does it for all where they aren't OptedOutFromCompaniesHouseUpdate and have a CompanyNumber. There's no filter for status on it unless we're setting OptedOutFromCompaniesHouseUpdate somewhere I haven't seen.

Since our existingOrganisation has been searched by CompanyNumber, it should have been picked up by this job, so we shouldn't need to.

malinaslevoaca1: Great then 👍 

You can see details about the background jobs in `BackgroundJobsApi.cs`. It seems that FetchCompaniesHouseDataJob runs every 5 minutes.

GenderPayGap.WebUI/Services/OrganisationService.cs
AlexJacksonDS: Again, looks like this is handled by one of the recurring jobs SetPresumedScopesJob (04:50 each day) that also doesn't have any filters for retired/deleted organisation to add new OrganisationScopes, so they should be there as presumed in/out.

malinaslevoaca1: Interesting!

malinaslevoaca1 (approved):

PR #356, opened 2021-03-24
GPG-NONE Content change fixes

AlexJacksonDS:
Content change fixes for:
ChooseSector.cshtml h1 text 'What type of employer do you want to add?'
No bullets + "more options" should be removed (remove user)
Scope confirmation header font
"Was your employer required to report for the 2019/20 reporting year?" to XXXX-XX
Declare scope confirmation via newly created organisations being broken (it was using a different layout page)

Also swapped a couple of places I noticed where the reporting year helper wasn't used

Checked these ones in Chome/Firefox (none were browser specific errors)

### DIFF
--- a/GenderPayGap.WebUI/Views/Organisation/ScopeDeclared.cshtml
+++ b/GenderPayGap.WebUI/Views/Organisation/ScopeDeclared.cshtml
@@ -1,18 +1,18 @@
 ﻿@{
     ViewBag.Title = "We need more information - Gender pay gap reporting service";
+    Layout = "~/Views/GovUkFrontend/GovUkFrontendLayout.cshtml";
 }
 @using GenderPayGap.Core
+@using GenderPayGap.Core.Helpers
 @using GenderPayGap.WebUI.Models.Shared.Patterns
-@using GovUkDesignSystem
-@using GovUkDesignSystem.GovUkDesignSystemComponents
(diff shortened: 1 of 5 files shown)


GenderPayGap.WebUI/Views/Organisation/ScopeDeclared.cshtml
```diff
@@ -1,18 +1,18 @@
 ﻿@{
     ViewBag.Title = "We need more information - Gender pay gap reporting service";
+    Layout = "~/Views/GovUkFrontend/GovUkFrontendLayout.cshtml";
```
malinaslevoaca1, line 3: Does this cause any issues? (i.e. does the page look the same?)

malinaslevoaca1 (approved):
Looks good 👍

GenderPayGap.WebUI/Views/Organisation/ScopeDeclared.cshtml
AlexJacksonDS: It's the layout page we use for most of the rest of the site, it did break everything, but I fixed them to make it look like the rest of the confirm pages

malinaslevoaca1: I see, thanks 😄

PR #358, opened 2021-03-25
GPG-651 Fixes for the Scope journeys

AlexJacksonDS:
Fixes for the links on the scope declared confirmation page and redirect at the end of the set in scope journey.

All these scope changing views/actions are probably a bit more complicated than they need to be, might be worth considering re-working them in the future

### DIFF
--- a/GenderPayGap.WebUI/Controllers/ScopeController.cs
+++ b/GenderPayGap.WebUI/Controllers/ScopeController.cs
@@ -143,7 +143,7 @@ namespace GenderPayGap.WebUI.Controllers
             
             SendScopeChangeEmails(organisation, viewModel.ReportingYear, currentSnapshotDate, ScopeStatuses.InScope);
 
-            return RedirectToAction("ManageOrganisationGet", "ManageOrganisations", new { encryptedOrganisationId = encryptedOrganisationId });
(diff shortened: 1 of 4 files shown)


GenderPayGap.WebUI/Controllers/ScopeController.cs
```diff
@@ -143,7 +143,7 @@ public IActionResult ConfirmInScopeAnswers(string encryptedOrganisationId, int r
             
             SendScopeChangeEmails(organisation, viewModel.ReportingYear, currentSnapshotDate, ScopeStatuses.InScope);
 
-            return RedirectToAction("ManageOrganisationGet", "ManageOrganisations", new { encryptedOrganisationId = encryptedOrganisationId });
+            return RedirectToAction("ScopeDeclared", "Organisation", new {id = encryptedOrganisationId});
```
malinaslevoaca1, line 146: Hmm, why do we need a new action?

AlexJacksonDS: The original journey didn't have this scope declared page, figured it was probably best to call the existing confirmation view in the Organisation area rather than add another new one in the Scope area. New action is because the existing view was only used as a return of the POST for the new org scope journey

malinaslevoaca1 (approved):

PR #359, opened 2021-03-25
GPG-651 Updated private sector deadline

malinaslevoaca1:


GenderPayGap.WebUI/Views/Organisation/ScopeDeclared.cshtml
```diff
@@ -50,8 +50,8 @@
                 </li>
                 <li>
                     private, voluntary and all other public authority employers must use
-                    a snapshot date of 31 March and report and publish gender pay gap
-                    information by 30 March of the following year.
+                    a snapshot date of 5 April and report and publish gender pay gap
```
AlexJacksonDS, line 53: I was going to do this like this instead, I think it would be easier

most public authority employers must use a snapshot date of @(Model.SnapshotDate.ToString("dd MMMM")) and 
report and publish gender pay gap information by @(Model.SnapshotDate.AddDays(-1).ToString("dd MMMM")) of the following year

malinaslevoaca1: but we use this view for both public and private organisations

AlexJacksonDS: Nevermind, re-reading it it's fine

AlexJacksonDS (approved):

PR #360, opened 2021-03-30
GPG-735 Added database integrity checks for the submission form data

malinaslevoaca1:
See https://technologyprogramme.atlassian.net/browse/GPG-735

- Added new integration checks
- Refactored existing views to use the breadcrumbs partial

### DIFF
--- a/GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
+++ b/GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
@@ -234,12 +234,11 @@ namespace GenderPayGap.WebUI.Controllers.Admin
             List<Return> invalidReturns =
                 dataRepository.GetAll<Return>()
                     .Where(r => r.Status == ReturnStatuses.Submitted)
-                    .AsEnumerable()
                     .Where(
-                        r => SumNotEqualTo100(r.FemaleLowerPayBand, r.MaleLowerPayBand)
-                             || SumNotEqualTo100(r.FemaleMiddlePayBand, r.MaleMiddlePayBand)
-                             || SumNotEqualTo100(r.FemaleUpperPayBand, r.MaleUpperPayBand)
-                             || SumNotEqualTo100(r.FemaleUpperQuartilePayBand, r.MaleUpperQuartilePayBand))
+                        r => r.FemaleLowerPayBand + r.MaleLowerPayBand != 100
+                             || r.FemaleMiddlePayBand + r.MaleMiddlePayBand != 100
+                             || r.FemaleUpperPayBand + r.MaleUpperPayBand != 100
+                             || r.FemaleUpperQuartilePayBand + r.MaleUpperQuartilePayBand != 100)
                     .ToList();
 
             return PartialView("ReturnsWithQuartersFiguresSumDifferentThan100", invalidReturns);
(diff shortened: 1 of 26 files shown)


GenderPayGap.WebUI/Helpers/IntegrationChecksHelper.cs
```diff
@@ -0,0 +1,28 @@
...
+    public static class IntegrationChecksHelper
+    {
+        public static Boolean SumNotEqualTo100(decimal a, decimal b)
+        {
+            return a + b != 100;
+        }
+
+        public static Boolean FigureGreaterThan100OrLessThan(decimal? a, decimal minValue)
+        {
+            return a > 100 || a < minValue;
```
Binney, line 14: what does this do if `a` doesn't have a value?

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
```diff
@@ -200,5 +201,192 @@ public IActionResult PrivateSectorOrganisationsWithAPublicSectorType()
...
+            return PartialView("ReturnsWithFiguresWithMoreThanOneDecimalPlace", invalidReturns);
+        }
+
+        [HttpGet("database-integrity-checks/returns-with-quarters-figures-sum-different-than-100")]
+        public IActionResult ReturnsWithQuartersFiguresSumDifferentThan100()
+        {
+            List<Return> invalidReturns =
+                dataRepository.GetAll<Return>()
+                    .Where(r => r.Status == ReturnStatuses.Submitted)
+                    .AsEnumerable()
```
Binney, line 237: I'm fairly sure (maybe worth checking!!) that this `AsEnumerable` call will download all the submitted returns into memory and further parse them there. This obviously poses issues if you've got an absolutely massive database (and is I assume why the other tests in this file except one don't use it, so Entity Framework automatically translates it the LINQ into SQL).

I think you ought to be able to some of these new tests entirely without `AsEnumerable` (see eg [here](https://stackoverflow.com/questions/25489331/get-the-sum-of-multiple-columns) for summing two fields, it's a bit like doing `SELECT (FemaleUpperPayBand + MaleUpperPayBand) AS sum WHERE sum != 100`. wdyt?

(It's a shame because I really like this style, it's really slick, I'm just interested if we can squeeze some extra performance out of it)

GenderPayGap.WebUI/Views/AdminDatabaseIntegrityChecks/ReturnsWithFiguresWithMoreThanOneDecimalPlace.cshtml
```diff
@@ -0,0 +1,33 @@
...
+    <partial name="~/Views/AdminDatabaseIntegrityChecks/DatabaseIntegrityChecksBreadcrumbs.cshtml" model="title"/>
+}
+
+<span class="govuk-caption-xl">Database integrity checks</span>
+<h1 class="govuk-heading-xl">@title</h1>
+
+<p class="govuk-body">
+    This shows submitted reports that contain figures with more than one decimal place.
+    <br/>
+    Please note that only one decimal is displayed everywhere on the service but you can see the bad data by downloading the CSV.
```
Binney, line 20: Great spot!

Binney (changes requested):
Looks good, main q is around efficiency, I should also caveat that I haven't read the original ticket list at all (ie I have not checked these are the right tests to be running) but I can if you'd like me to, just let me know :)

GenderPayGap.WebUI/Helpers/IntegrationChecksHelper.cs
malinaslevoaca1: it returns false

GenderPayGap.WebUI/Controllers/Admin/AdminDatabaseIntegrityChecksController.cs
malinaslevoaca1: Good point! I changed the queries so that they can be translated to SQL. There were 2 that cannot be translated so I kept them as they were (figure has more than 1 decimal place and company link is invalid)

Binney (approved):
Awesomeeeeeeee :+1:

PR #364, opened 2021-04-01
GPG-721 Default order by order added to comparison basket

AlexJacksonDS:
Changed the results from the compare to sort by the passed in list of encrypted organisation IDs by default so that they are sorted in the order that they are added to the comparison basket rather than by the numeric organisation IDs from the database.


malinaslevoaca1 (approved):
👍

Binney (approved):

PR #367, opened 2021-04-06
GPG-653 Content changes for the Search and Compare screens

AlexJacksonDS:
Content changes for the Search and Compare screens, a mixture of layout and wording changes in this one.

After Jacob's latest comment on the ticket, and discussing with @malinaslevoaca1, we are going to do this ticket exactly as required by the pdf of content changes and then raise a second ticket for the layout changes for converting the green basket banner on the other pages (search and employer details) separately if we want to make that change. The second ticket should be relatively simple building upon the work from this ticket, changing the partial used for the basket and making sure the layout is still fine.

This also covers GPG-720 as the wording that had a weird layout has been changed and now should look OK.


GenderPayGap.WebUI/Scripts/Application/LiveSearch.js
```diff
@@ -118,10 +118,12 @@
 
     LiveSearch.prototype.showLoadingIndicator = function showLoadingIndicator() {
         this.$loadingBlock.text('Loading...');
```
malinaslevoaca1, line 120: When is this displayed?

GenderPayGap.WebUI/Views/Compare/CompareEmployers.cshtml
```diff
@@ -2,89 +2,82 @@
...
+                    </span>
+                </summary>
+                <div class="govuk-details__text">
+                    <partial name="AboutQuartiles" />
+                </div>
             </details>
         </div>
-        <div class="column-full">
-            
+        <div class="govuk-grid-column-full">
```
malinaslevoaca1, line 54: why do we have this empty div?

GenderPayGap.WebUI/Views/Compare/ComparisonBasket.cshtml
```diff
@@ -11,7 +11,7 @@
 }
 
 <div id="comparison-basket"
-     class="no-print show-empty js-stick-at-top-when-scrolling @(compareCount < 1 ? "visuallyhidden" : "")"
+     class="comparison-basket no-print show-empty js-stick-at-top-when-scrolling @(compareCount < 1 ? "visuallyhidden" : "")"
```
malinaslevoaca1, line 14: Do we still need ` @(compareCount < 1 ? "visuallyhidden" : "")`?

malinaslevoaca1 (changes requested):
I had a look locally and noticed that there are some things left to be done (page 2 in the PDF): 
- Employers font size not increased
- Filters columns not aligned with the divider line

GenderPayGap.WebUI/Views/Compare/ComparisonBasket.cshtml
AlexJacksonDS: Yes, this hides the green basket banner on the search/employer pages when there is nothing in the basket. This functionality needs to stay the same until(if) we change from the green banner basket on those pages

GenderPayGap.WebUI/Scripts/Application/LiveSearch.js
AlexJacksonDS: Not sure whether it is ever shown with the 'Loading...' variant. It may be if the search functionality is being slow. The '&nbsp;' variant from line 125 is definitely show, and causes misalignment without the added this.$loadingBlock.hide(); on line 126. Currently it is here:
image

AlexJacksonDS:
> I had a look locally and noticed that there are some things left to be done (page 2 in the PDF):
> 
> * Employers font size not increased
> * Filters column not aligned with the divider line

Updated the font size.
The filter columns not being aligned, I'm not sure what you mean here. After the changes it looks like this:
image

malinaslevoaca1:
> > I had a look locally and noticed that there are some things left to be done (page 2 in the PDF):
> > 
> > * Employers font size not increased
> > * Filters column not aligned with the divider line
> 
> Updated the font size.
> The filter columns not being aligned, I'm not sure what you mean here. After the changes it looks like this:
> image

There is a comment: "This column. block sits on same horizontal as the dividing line". It should look like this:
image

AlexJacksonDS:
Trying to pull it into the partial as discussed causes this so reverting to alignment at the top rather than spending more time on it.
image

malinaslevoaca1 (approved):
Nice!

PR #370, opened 2021-04-08
Gpg 744 updating robots.txt

AlexJacksonDS:
GPG-744: Added the SEMrush user agents to the robots.txt file

GPG-584: 
- Went through and removed the disallowed endpoints that are behind authorisation and a couple that no longer exist.
- /register/ is still there as I've just added the /Register/about-you endpoint to redirect to the new create account page in GPG-742.
- The remaining disallowed endpoints under /viewing/ are all the ones that relate to functionality like the add to compare buttons etc. that should probably remain disallowed as they aren't actual pages. Added a couple of extra ones I found in the viewing controller that weren't previously included in the list of disallowed endpoints, but probably should have been as they fall into the same category.

I think this is still the correct approach to take in terms of allow/disallow, rather than having to specify all the public pages that we do want to get crawled individually.

### DIFF
--- a/GenderPayGap.WebUI/wwwroot/robots.txt
+++ b/GenderPayGap.WebUI/wwwroot/robots.txt
@@ -1,10 +1,5 @@
 User-agent: *
-Disallow: /admin/
-Disallow: /home/
 Disallow: /register/
-Disallow: /submit/
-Disallow: /test/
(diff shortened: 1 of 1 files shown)


Binney (approved):
Sounds very sensible to me :)

PR #373, opened 2021-04-27
GPG-655: Add Filter for anonymised users

xonde:


GenderPayGap.WebUI/Services/EmailSendingServiceHelpers.cs
```diff
@@ -11,7 +11,7 @@ public static void SendUserAddedEmailToExistingUsers(Organisation organisation,
         {
             IEnumerable<string> emailAddressesForOrganisation = organisation.UserOrganisations
                 .Select(uo => uo.User.EmailAddress)
-                .Where(ea => ea != addedUser.EmailAddress);
+                .Where(ea => ea != addedUser.EmailAddress && ea != "anonymised");
```
malinaslevoaca1, line 14: We should use `HasBeenAnonymised` instead.

malinaslevoaca1 (changes requested):
I know most (if not all) of the failures come from new user registrations so it's good to have this filter but I think we should also check if the email address is anonymised before sending any email (as in [here](https://github.com/cabinetoffice/gender-pay-gap/blob/d293c0d35f2a7253901bcb91f62be536b8f816bf/GenderPayGap.WebUI/Services/EmailSendingService.cs#L424))

xonde:
> I know most (if not all) of the failures come from new user registrations so it's good to have this filter but I think we should also check if the email address is anonymised before sending any email (as in [here](https://github.com/cabinetoffice/gender-pay-gap/blob/d293c0d35f2a7253901bcb91f62be536b8f816bf/GenderPayGap.WebUI/Services/EmailSendingService.cs#L424))

Ah yes, I initially made the change here but I wasn't sure if that was what ticket wanted, will revert it to that now

malinaslevoaca1:
> > I know most (if not all) of the failures come from new user registrations so it's good to have this filter but I think we should also check if the email address is anonymised before sending any email (as in [here](https://github.com/cabinetoffice/gender-pay-gap/blob/d293c0d35f2a7253901bcb91f62be536b8f816bf/GenderPayGap.WebUI/Services/EmailSendingService.cs#L424))
> 
> Ah yes, I initially made the change here but I wasn't sure if that was what ticket wanted, will revert it to that now

@xonde Well, we could keep this filter but it should be there as well

malinaslevoaca1 (approved):
Nice

PR #375, opened 2021-04-28
GPG-697: Removed Survey links

xonde:
Removed all links to user survey (including pages we don't use)


malinaslevoaca1 (approved):
I think we should remove the files as well. If needed to be readded (I doubt that) they could be found in the git history

dulciejackson (approved):
Looks good! :)

PR #381, opened 2021-05-11
Fix ErrorController default action bug and redirect to /error if Ajax…

malinaslevoaca1:
… request fails

See ticket [GPG-763](https://technologyprogramme.atlassian.net/browse/GPG-763)

Issues
1. /error/errorCode request fails with 502 Bad Gateway if the code is not a valid status code.
2. User is redirected to /error/1146 if Ajax request fails

Acceptance criteria
1. /error/errorCode returns the "Something's gone wrong" page (500 error) if errorCode is not a valid HTTP status code
2. User is redirected to /error if Ajax request fails


Lucas-Fernandes-Nogueira (approved):
It all looks good to me 😁.

PR #382, opened 2021-05-13
GPG-711 reminder email fixes

malinaslevoaca1:
**Issue:** Only the latest reporting year (2021) was considered when sending the reminder emails.

**Fix:** Select all the reporting years that have upcoming deadlines.

I've also added a filter for the anonymised users.

I tested this on the load test environment and emails are sent. You can see an example [here](https://technologyprogramme.atlassian.net/browse/GPG-711?focusedCommentId=66418).


Lucas-Fernandes-Nogueira (approved):
It all looks good to me 😊.
Maybe it would be worth updating some of the function names to reflect the fact that they now take year as an argument as well?

Lucas-Fernandes-Nogueira (approved):
It looks good to me 😁.

PR #383, opened 2021-05-13
GPG-712 Late submission fix

malinaslevoaca1:
**Issue:**
After removing 2020 from `ReportingStartYearsToExcludeFromLateFlagEnforcement` reports were marked as late (even though the deadline for 2020 was in October)

**Fix:**
Updated the late reports logic: check whether the deadline has already passed or not.

I've already tested this on the load test environment. [Here](https://technologyprogramme.atlassian.net/browse/GPG-712) are all the things that I've checked.


Lucas-Fernandes-Nogueira (approved):
It all looks good 👍.

PR #386, opened 2021-05-22
GPG-761: remove duplicate filter title

Lucas-Fernandes-Nogueira:
T- I went to /viewing/search-results?t=1&search=&orderBy=relevance and checked that the filter titles don't appear twice. I also tested with NVA screen reader and it read the filter titles correctly.
R- Very low risks. Maybe we should remove the legend, since it might be unnecessary.
D- No need.

### DIFF
--- a/GenderPayGap.WebUI/Content/Site.css
+++ b/GenderPayGap.WebUI/Content/Site.css
@@ -663,6 +663,10 @@ input[type=submit] {
     height: 20px;
 }
 
+.finder .govuk-option-select .options-container .option-select-legend {
+    display: none;
+}
+
 .js-enabled .finder .govuk-option-select {
     padding: 0 0 1px 0;
     /* styles for collapsibleness. .js-collapsible is added by the javascript if the browser is not ie6/7 in which case these don't collapse */
--- a/GenderPayGap.WebUI/Content/Site.css
+++ b/GenderPayGap.WebUI/Content/Site.css
@@ -663,10 +663,6 @@ input[type=submit] {
     height: 20px;
 }
 
-.finder .govuk-option-select .options-container .option-select-legend {
-    display: none;
-}
-
 .js-enabled .finder .govuk-option-select {
     padding: 0 0 1px 0;
     /* styles for collapsibleness. .js-collapsible is added by the javascript if the browser is not ie6/7 in which case these don't collapse */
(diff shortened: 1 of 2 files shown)


Lucas-Fernandes-Nogueira:
Maybe instead of hiding the filter legend, we should remove it altogether? I tried finding references to it in the code, and it seems like the legend isn't being used. I also deleted it in the browser and tested with a screen reader, and it still read the filter titles correctly.

GenderPayGap.WebUI/Views/Viewing/Finder/Parts/Facets/OptionSelect.cshtml
```diff
@@ -12,7 +12,7 @@
     <div class="options-container" id="@(Model.Id)" aria-label="@(Model.Label) filter" style="max-height: @(maxHeight)">
         <div class="js-auto-height-inner">
             <fieldset>
-                <legend>@(Model.Label)</legend>
+                <legend class="option-select-legend">@(Model.Label)</legend>
```
malinaslevoaca1, line 15: we could use `class="govuk-visually-hidden"` instead

GenderPayGap.WebUI/Content/Site.css
```diff
@@ -663,6 +663,10 @@ input[type=submit] {
     height: 20px;
 }
 
+.finder .govuk-option-select .options-container .option-select-legend {
+    display: none;
```
malinaslevoaca1, line 667: `display: none;` hides the content from the screen readers as well so that might be the reason why you've seen no difference. You can read [here](https://webaim.org/techniques/css/invisiblecontent/) about different techniques for hiding the content. 
It also seems that some screen readers ignore the `legend` tags. However. even if it's optional it's best practice to have a legend for each field set to describe the field set data.

malinaslevoaca1 (changes requested):
We should use `govuk-visually-hidden` class from [govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/helpers/_visually-hidden.scss) instead.

GenderPayGap.WebUI/Views/Viewing/Finder/Parts/Facets/OptionSelect.cshtml
Lucas-Fernandes-Nogueira: That's way more sensible! Thanks 😊.

GenderPayGap.WebUI/Content/Site.css
Lucas-Fernandes-Nogueira: This is really helpful! Thanks Malina :) .

Lucas-Fernandes-Nogueira:
I've tested with the screen reader again after the final changes, and this is the reader output:

> Visited link "Clear all filters" [press down arrow]
> 
> Heading level to search filters [press down arrow]
> 
> Employer size [press down arrow]
> 
> Grouping Employer size [press down arrow]
> 
> Checkbox not checked Not Provided, Checkbox not checked Less than 250, ... [press down arrow]
> 
This looks fine to me, but please let me know if the screen reader should be saying something different 😃.

malinaslevoaca1 (approved):
Looks good to me! 👍

PR #389, opened 2021-05-25
GPG-771: fix non expandable filters bug.

Lucas-Fernandes-Nogueira:
T - Went to the viewing page. Checked that the filters were expandable. Selected a few filters, and checked that the counter under the filter group title were udpated.
R - Very low. This is code that already existed but had been wrongly removed (by me 😅)
D - no need.

### DIFF
--- a/GenderPayGap.WebUI/Views/Viewing/Finder/SearchResults.cshtml
+++ b/GenderPayGap.WebUI/Views/Viewing/Finder/SearchResults.cshtml
@@ -52,6 +52,12 @@
 
             //Keep comparison basket within viewport
             GOVUK.stickAtTopWhenScrolling.init();
+
+            // Instantiate an option select for each one found on the page
+            // This will make each filter group expandable and add a counter of selected filters
+            $('#FinderForm .govuk-option-select').map(function () {
+                new GOVUK.OptionSelect({ $el: $(this) });
+            });
             
             let serializedModel = @Json.Serialize(Model);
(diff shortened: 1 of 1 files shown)


GenderPayGap.WebUI/Views/Viewing/Finder/SearchResults.cshtml
```diff
@@ -52,6 +52,12 @@
 
             //Keep comparison basket within viewport
             GOVUK.stickAtTopWhenScrolling.init();
+
+            // Instantiate an option select for each one found on the page
+            // This will make each filter group expandable and add a counter of selected filters
+            $('#FinderForm .govuk-option-select').map(function () {
+                new GOVUK.OptionSelect({ $el: $(this) });
+            });
```
Lucas-Fernandes-Nogueira, line 60: This is what this code used to look like, before I removed it:
```
            // Instantiate an option select for each one found on the page
            var filters = $('#FinderForm .govuk-option-select').map(function () {
                return new GOVUK.OptionSelect({ $el: $(this) });
            });

```
I don't like the approach here tbh. At the time I deleted these lines of code because the filters variable wasn't used and because I wasn't expecting a map to be able to change its elements. I've now added a comment explaining a bit better why we need these lines of code, but I still think this approach is a bit confusing.

malinaslevoaca1 (approved):
Thanks for fixing this!

PR #390, opened 2021-05-27
GPG-752 Fixed csv negative values bug

malinaslevoaca1:
See ticket [GPG-768](https://technologyprogramme.atlassian.net/browse/GPG-768)

We have the following CSV files:
- Admin downloads
- Admin feedback download
- Comparison results
- Public facing files

I excluded '-' from the injection characters as the injection function cannot be configured by us. However, we can add a custom type converter. Here all the values that start with '-' and are not decimals (we have only decimal figures) are prepended with a tab.

All the CSV files should be sanitized. Therefore, I added a custom `CsvWriter` and used it everywhere.

Added tests. Apart from that I've also checked the AC pass for all the files apart from the public facing ones locally. The public facing files are built by a background job - we don't run them locally. I will make sure I'll test this once deployed to the dev environment.

### DIFF
--- a/GenderPayGap.Core/Helpers/CsvWriter.cs
+++ b/GenderPayGap.Core/Helpers/CsvWriter.cs
@@ -10,6 +10,13 @@ namespace GenderPayGap.Core.Helpers
     public static class CsvWriter
     {
 
+        /*
+         * The default injection characters are: '=', '@', '+', '-'
+         * We ignore '-' because we don't want the negative figures to be sanitized (i.e., prepended with a tab character)
+         * We sanitize manually the other fields that start with '-' using the CustomConverter
+         */
+        private static readonly char[] InjectionCharacters = {'=', '@', '+'};
+
         public static T Write<T>(Func<MemoryStream, StreamReader, StreamWriter, CsvHelper.CsvWriter, T> write)
         {
             var config = new CsvConfiguration(CultureInfo.CurrentCulture)
@@ -25,13 +32,6 @@ namespace GenderPayGap.Core.Helpers
                 return write(memoryStream, streamReader, streamWriter, csvWriter);
             }
         }
-
-        /*
-         * The default injection characters are: '=', '@', '+', '-'
-         * We ignore '-' because we don't want the negative figures to be sanitized (i.e., prepended with a tab character)
-         * We sanitize manually the other fields that start with '-' using the CustomConverter
-         */
-        private static readonly char[] InjectionCharacters = {'=', '@', '+'};
 
         private class CustomConverter : DefaultTypeConverter
         {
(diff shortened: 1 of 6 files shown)


GenderPayGap.Core/Helpers/CsvWriter.cs
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    return '\t' + value.ToString();
+                }
+
+                return base.ConvertToString(value, row, memberMapData);
+            }
+
+            private bool HasToBeSanitized(object value)
+            {
+                return value is string s && !decimal.TryParse(s, out _) && s.StartsWith('-');
```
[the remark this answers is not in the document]
malinaslevoaca1: > Will this also try to sanitise a negative integer?

Integers (and decimals) won't be sanitized.

> Do we also want to sanitise a positive number that starts with a +?

Any value that starts with a + is going to be sanitised. I don't think we should change that because, as you said, the positive numbers don't have + prepended.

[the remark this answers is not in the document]
Lucas-Fernandes-Nogueira: > Any value that starts with a + is going to be sanitised. I don't think we should change that because, as you said, the positive numbers don't have + prepended.
I accept it then 😀.

GenderPayGap.Core/Helpers/CsvWriter.cs
```diff
@@ -0,0 +1,59 @@
...
+                return write(memoryStream, streamReader, streamWriter, csvWriter);
+            }
+        }
+
+        /*
+         * The default injection characters are: '=', '@', '+', '-'
+         * We ignore '-' because we don't want the negative figures to be sanitized (i.e., prepended with a tab character)
+         * We sanitize manually the other fields that start with '-' using the CustomConverter
+         */
+        private static readonly char[] InjectionCharacters = {'=', '@', '+'};
```
Lucas-Fernandes-Nogueira, line 34: Should this be at the top (just because the Write function uses it)? I understand maybe the convention is different?
Fine if you don't agree, I've approved the changes 😁.
(sorry I'm only suggesting this now)

Lucas-Fernandes-Nogueira (dismissed):

GenderPayGap.Core/Helpers/CsvWriter.cs
malinaslevoaca1: @Lucas-Fernandes-Nogueira
I usually order the items by their access:
- public
- protected
- private

However, as I haven't written that much C# code, I've googled that and found [this Stackoverflow answer](https://stackoverflow.com/questions/150479/order-of-items-in-classes-fields-properties-constructors-methods)(the first answer) that I find useful. I will update this class accordingly. Thanks for flagging!

Lucas-Fernandes-Nogueira: Thanks for sharing that link, it's really helpful!

Lucas-Fernandes-Nogueira (approved):

PR #391, opened 2021-05-31
GPG-751 Fixed incorrect date bug

malinaslevoaca1:
The majority of the database tables contain a column for the created date. None of the dates contain the timezone.
 
`VirtualDateTime.Now` uses the system timezone - As this bug hasn't been reported until now, the system timezone must have been "Europe/London" initially. However, now the system timezone is "UTC".
As we can't consider all of the dates (from the db) to be in UTC format, we should change the system timezone. Added an environment variable for setting this.

We use `VirtualDateTime.Now` for all the entries apart from the audit logs and the feedback ones, for these the default value is now() that uses the database timezone. 
We should use "VirtualDateTime.Now" everywhere for consistency.

**Tests:** I deployed this build to the dev environment to test the manifest file change and everything works as expected, the system timezone being Europe/London.

### DIFF
--- a/GenderPayGap.Core/Extensions/VirtualDateTime.cs
+++ b/GenderPayGap.Core/Extensions/VirtualDateTime.cs
@@ -7,6 +7,7 @@ namespace GenderPayGap.Extensions
 
         private static TimeSpan offsetCurrentDateTimeForSite = TimeSpan.Zero;
 
+        // This uses the system timezone
         public static DateTime Now => DateTime.Now.Add(offsetCurrentDateTimeForSite);
 
         public static DateTime UtcNow => DateTime.UtcNow.Add(offsetCurrentDateTimeForSite);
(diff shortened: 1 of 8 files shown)


Lucas-Fernandes-Nogueira (approved):
It all looks good to me 😁.

Do we need to go through the database and update all the datetime values to Europe/London format?

malinaslevoaca1:
> It all looks good to me 😁.
> 
> Do we need to go through the database and update all the datetime values to Europe/London format?

Unfortunately, we can't do that as the database records have the date without timezone

Lucas-Fernandes-Nogueira:
> > It all looks good to me 😁.
> > Do we need to go through the database and update all the datetime values to Europe/London format?
> 
> Unfortunately, we can't do that as the database records have the date without timezone

You mean we don't know which dates are in UTC and which are in Europe/London?

malinaslevoaca1:
> > > It all looks good to me 😁.
> > > Do we need to go through the database and update all the datetime values to Europe/London format?
> > 
> > 
> > Unfortunately, we can't do that as the database records have the date without timezone
> 
> You mean we don't know which dates are in UTC and which are in Europe/London?

Exactly

Lucas-Fernandes-Nogueira:
> > > > It all looks good to me 😁.
> > > > Do we need to go through the database and update all the datetime values to Europe/London format?
> > > 
> > > 
> > > Unfortunately, we can't do that as the database records have the date without timezone
> > 
> > 
> > You mean we don't know which dates are in UTC and which are in Europe/London?
> 
> Exactly

Sad, but unavoidable 👍.

PR #392, opened 2021-05-31
GPG-770 Added test for reporting year excluded from late flag enforcement

malinaslevoaca1:
See ticket [GPG-770](https://technologyprogramme.atlassian.net/browse/GPG-770)

I checked all the AC on the dev environment.
We already have tests for these. I added one test that was missing - check if the report is marked as late if the reporting year is excluded from late flag enforcement.

### DIFF
--- a/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
+++ b/GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using GenderPayGap.Core;
 using GenderPayGap.Core.Classes;
 using GenderPayGap.Core.Helpers;
@@ -12,38 +14,32 @@ namespace GenderPayGap.Database.ReturnTests
     public class IsLateSubmissionTests
     {
 
+        private static List<int> reportingStartYearsToExcludeFromLateFlagEnforcement = Global.ReportingStartYearsToExcludeFromLateFlagEnforcement;
+
         [TestCase(SectorTypes.Public, ScopeStatuses.InScope)]
         [TestCase(SectorTypes.Public, ScopeStatuses.PresumedInScope)]
         [TestCase(SectorTypes.Private, ScopeStatuses.InScope)]
         [TestCase(SectorTypes.Private, ScopeStatuses.PresumedInScope)]
         public void Is_True_When_ModifiedDate_Is_Late_And_InScope(SectorTypes sector, ScopeStatuses scopeStatus)
         {
-            var totalYearOffsets = 4;
-            for (var yearOffset = 0; yearOffset < totalYearOffsets; yearOffset++)
-            {
-                // Arrange 
-                int testYear = VirtualDateTime.Now.Year - yearOffset;
-                if (Global.ReportingStartYearsToExcludeFromLateFlagEnforcement.Contains(testYear))
-                {
-                    continue;
-                }
-                DateTime snapshotDate = sector.GetAccountingStartDate(testYear);
-                DateTime modifiedDate = ReportingYearsHelper.GetDeadlineForAccountingDate(snapshotDate).AddDays(2);
+            // Arrange 
+            int testYear = GetRandomReportingYear(ignoreYearsExcludedFromLateFlagEnforcement: true);
+            DateTime snapshotDate = sector.GetAccountingStartDate(testYear);
+            DateTime modifiedDate = ReportingYearsHelper.GetDeadlineForAccountingDate(snapshotDate).AddDays(2);
 
-                Organisation testOrganisation = sector == SectorTypes.Private
-                    ? OrganisationHelper.GetPrivateOrganisation()
-                    : OrganisationHelper.GetPublicOrganisation();
+            Organisation testOrganisation = sector == SectorTypes.Private
+                ? OrganisationHelper.GetPrivateOrganisation()
+                : OrganisationHelper.GetPublicOrganisation();
 
-                OrganisationScope testScope = ScopeHelper.CreateScope(scopeStatus, snapshotDate);
+            OrganisationScope testScope = ScopeHelper.CreateScope(scopeStatus, snapshotDate);
 
-                Return testReturn = ReturnHelper.CreateLateReturn(testOrganisation, snapshotDate, modifiedDate, testScope);
+            Return testReturn = ReturnHelper.CreateLateReturn(testOrganisation, snapshotDate, modifiedDate, testScope);
 
-                // Act
-                bool actual = testReturn.IsLateSubmission;
+            // Act
+            bool actual = testReturn.IsLateSubmission;
 
-                // Assert
-                Assert.AreEqual(true, actual);
-            }
+            // Assert
+            Assert.AreEqual(true, actual);
         }
 
         [TestCase(SectorTypes.Public, ScopeStatuses.InScope)]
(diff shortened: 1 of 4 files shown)


GenderPayGap.WebUI/appsettings.LOCAL.json
```diff
@@ -2,5 +2,6 @@
   "DaysToKeepBackupFiles": 0,
   "GoogleAnalyticsAccountId": "UA-93591336-1",
   "FeatureFlagNewReportingJourney": "true",
-  "FeatureFlagNewManageOrganisationsJourney": "true"
+  "FeatureFlagNewManageOrganisationsJourney": "true",
+  "ReportingStartYearsToExcludeFromLateFlagEnforcement": [2019]
```
Lucas-Fernandes-Nogueira, line 6: Did this use to be [2019,2020]? Where was this set?

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
```diff
@@ -110,5 +114,36 @@ public void Is_False_When_ModifiedDate_Is_Late_And_OutOfScope(SectorTypes sector
...
 
+        [TestCase(SectorTypes.Public, ScopeStatuses.InScope)]
+        [TestCase(SectorTypes.Public, ScopeStatuses.PresumedInScope)]
+        [TestCase(SectorTypes.Private, ScopeStatuses.InScope)]
+        [TestCase(SectorTypes.Private, ScopeStatuses.PresumedInScope)]
+        [TestCase(SectorTypes.Public, ScopeStatuses.OutOfScope)]
+        [TestCase(SectorTypes.Public, ScopeStatuses.PresumedOutOfScope)]
+        [TestCase(SectorTypes.Private, ScopeStatuses.OutOfScope)]
+        [TestCase(SectorTypes.Private, ScopeStatuses.PresumedOutOfScope)]
+        public void IsFalse_When_ModifiedDate_Is_Late_And_ExcludeFromLateFlagEnforcement_Year(SectorTypes sector, ScopeStatuses scopeStatus)
```
Lucas-Fernandes-Nogueira, line 125: Missing an underscore between Is and False.

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
```diff
@@ -110,5 +114,36 @@ public void Is_False_When_ModifiedDate_Is_Late_And_OutOfScope(SectorTypes sector
...
+                Organisation testOrganisation = sector == SectorTypes.Private
+                    ? OrganisationHelper.GetPrivateOrganisation()
+                    : OrganisationHelper.GetPublicOrganisation();
+
+                OrganisationScope testScope = ScopeHelper.CreateScope(scopeStatus, snapshotDate);
+
+                Return testReturn = ReturnHelper.CreateLateReturn(testOrganisation, snapshotDate, modifiedDate, testScope);
+
+                // Act
+                bool actual = testReturn.IsLateSubmission;
```
Lucas-Fernandes-Nogueira, line 142: I assume this line is here because we're trying to split the test into three. Would it be better if we merge this with the following line, so:
Assert.AreEqual(false, testReturn.IsLateSubmission).
Just because I wasn't sure what we meant by "actual".

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
```diff
@@ -110,5 +114,36 @@ public void Is_False_When_ModifiedDate_Is_Late_And_OutOfScope(SectorTypes sector
...
+        [TestCase(SectorTypes.Public, ScopeStatuses.PresumedInScope)]
+        [TestCase(SectorTypes.Private, ScopeStatuses.InScope)]
+        [TestCase(SectorTypes.Private, ScopeStatuses.PresumedInScope)]
+        [TestCase(SectorTypes.Public, ScopeStatuses.OutOfScope)]
+        [TestCase(SectorTypes.Public, ScopeStatuses.PresumedOutOfScope)]
+        [TestCase(SectorTypes.Private, ScopeStatuses.OutOfScope)]
+        [TestCase(SectorTypes.Private, ScopeStatuses.PresumedOutOfScope)]
+        public void IsFalse_When_ModifiedDate_Is_Late_And_ExcludeFromLateFlagEnforcement_Year(SectorTypes sector, ScopeStatuses scopeStatus)
+        {
+            foreach (int testYear in Global.ReportingStartYearsToExcludeFromLateFlagEnforcement)
```
Lucas-Fernandes-Nogueira, line 127: When the test fails, does it give enough information about which iteration of the foreach loop failed? Would it be better to implement this as a TestCase, like you did for the SectorType and ScopeStatus? Just so that we don't test two things in a single test.

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
```diff
@@ -23,6 +23,10 @@ public void Is_True_When_ModifiedDate_Is_Late_And_InScope(SectorTypes sector, Sc
             {
                 // Arrange 
                 int testYear = VirtualDateTime.Now.Year - yearOffset;
+                if (Global.ReportingStartYearsToExcludeFromLateFlagEnforcement.Contains(testYear))
+                {
+                    continue;
+                }
```
Lucas-Fernandes-Nogueira, line 29: I don't understand what we're trying to achieve here? Could you explain please 😄?

Lucas-Fernandes-Nogueira (changes requested):

GenderPayGap.WebUI/appsettings.LOCAL.json
malinaslevoaca1: No, it wasn't set locally; it defaults to [].
On the other environments this is set during the release

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
malinaslevoaca1: Yes, that's right. When running a test we have the expected result and the actual result. This applies to all the tests and that's why this variable was named like that. I'd rather keep it as it is cause the test looks clearer this way.

GenderPayGap.WebUI/appsettings.LOCAL.json
Lucas-Fernandes-Nogueira: I didn't notice this file was for local development. That makes sense, thanks 😄.

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
Lucas-Fernandes-Nogueira: Great explanation! I agree 👍.

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
malinaslevoaca1: Not sure if possible (I'll have a look). But the array contains a single value anyways

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
malinaslevoaca1: To be honest I don't really understand the reason why this test is implemented this way and I should probably refactor it.
**What it does:** It checks if a report is marked as late if submitted after the deadline. We need to test that for a reporting year (but why do we need 4 years?? - no idea).

**Why I added these lines:** Now `ReportingStartYearsToExcludeFromLateFlagEnforcement` is configured as well. Therefore, if one of the years is excluded we should skip it. Otherwise, the test will fail as reports should NOT be marked as late if the year is excluded from late flag enforcement.

**What I should have done:** Update the test to check this behavior for a SINGLE year that's not excluded from late flag enforcement.

Lucas-Fernandes-Nogueira: Sorry, I had completely forgotten about what the continue statement is for 😅.
Your explanation is quite good, thanks!

I'm also a bit confused why they test for 4 years. I don't think it's bad to test for more than one year, but again, I think this should be done as a TestCase, instead of using a for loop (I know this is possibly in python, no idea about C#).

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
Lucas-Fernandes-Nogueira, line 142: Another thing I forgot to mention: All the tests repeat these lines. Is it possible to commonise at least some this code?

Lucas-Fernandes-Nogueira (changes requested):

GenderPayGap.UnitTests/GenderPayGap.WebUI.Tests/DatabaseTests/Return/ReturnTests.IsLateSubmission.cs
malinaslevoaca1: ah sorry. `continue;` skips the current iteration of the loop and forces the next one
`break;` also skips the current iteration but it exits the loop instead of going to the next iteration

Lucas-Fernandes-Nogueira (approved):
It all looks good to me now 😁.

PR #393, opened 2021-06-02
GPG-766: add employer id to csv

Lucas-Fernandes-Nogueira:
This covers [GPG-760](https://technologyprogramme.atlassian.net/browse/GPG-766), which is to add the organisation ID to the CSV file in order to allow researchers to link an organisation over time.

T - Deployed on Dev and did the following: 
1. Logged in as an admin. 
2. Opened [hangfire](https://gender-pay-gap-dev.london.cloudapps.digital/admin/hangfire) 
3. Triggered the UPDATE_PUBLIC_FACING_DOWNLOAD_FILES_JOB. 
4. Went to [downloads page](https://gender-pay-gap-dev.london.cloudapps.digital/viewing/download) 
5. Downloaded all the csv files. 
6. Checked that the second column on each file is titled "EmployerId" 
7. Checked that the employer ids were present and were as expected 
8. Clicked "Search and compare" in the header 
9. Selected a few companies for comparison 
10. Went to the compare page 
11. Clicked "Download data (CSV)" 
12. Checked that the CSV does not contain the EmployerId column.

R - Very low risk. My code assumes the EmployerId always exists. This is a primary key, so that should be ok.
D -  not necessary.

### DIFF
--- a/GenderPayGap.Core/Models/DownloadResult.cs
+++ b/GenderPayGap.Core/Models/DownloadResult.cs
@@ -6,7 +6,7 @@ namespace GenderPayGap.Core.Models
     {
 
         public string EmployerName { get; set; }
-        public long OrganisationId { get; set; }
+        public long EmployerId { get; set; }
         public string Address { get; set; }
         public string CompanyNumber { get; set; }
         public string SicCodes { get; set; }
--- a/GenderPayGap.Core/Models/DownloadResult.cs
+++ b/GenderPayGap.Core/Models/DownloadResult.cs
@@ -6,6 +6,7 @@ namespace GenderPayGap.Core.Models
     {
 
         public string EmployerName { get; set; }
+        public long OrganisationId { get; set; }
         public string Address { get; set; }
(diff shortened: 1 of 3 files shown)


GenderPayGap.Core/Models/DownloadResult.cs
```diff
@@ -6,6 +6,7 @@ public class DownloadResult
     {
 
         public string EmployerName { get; set; }
+        public long OrganisationId { get; set; }
```
malinaslevoaca1, line 9: Recently all the user facing wording has been updated - "organisation" being changed to "employer". Therefore, I think we should rename this field

malinaslevoaca1 (commented):
Nice!

GenderPayGap.Core/Models/DownloadResult.cs
Lucas-Fernandes-Nogueira: Sure, will do that! Thanks 🙏.

malinaslevoaca1 (approved):

PR #395, opened 2021-06-11
GPG-775 CSV fields details

malinaslevoaca1:
I've deployed this build to dev. You can see how it looks like [here](https://gender-pay-gap-dev.london.cloudapps.digital/viewing/download). Tested on mobile as well.

The design was reviewed by Ariana.


GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/UpdatePublicFacingDownloadFilesJob.cs
```diff
@@ -94,6 +95,11 @@ private void SaveCsvFile(IEnumerable records, string relativeFilePath)
                 {
                     csvWriter.Configuration.ShouldQuote = (s, context) => true;
                     csvWriter.Configuration.TrimOptions = TrimOptions.InsideQuotes;
+
+                    var options = new TypeConverterOptions {Formats = new[] {"yyyy/MM/dd HH:mm:ss"}};
```
malinaslevoaca1, line 99: This was the format the dates were using but I added this here just in case the default format is changed.

GenderPayGap.WebUI/Controllers/ViewingController.cs
```diff
@@ -212,7 +212,7 @@ public IActionResult SuggestEmployerNameJs(string search)
         [HttpGet("download")]
         public IActionResult Download()
         {
-            return View("Download");
+            return View("Download/Download");
```
Lucas-Fernandes-Nogueira, line 215: Why do we need this change?

Lucas-Fernandes-Nogueira (approved):
This all looks good to me 😁.
(left a comment with a question... the answer is probably obvious, but it's Friday afternoon, so...)

GenderPayGap.WebUI/Controllers/ViewingController.cs
malinaslevoaca1: Since now we have multiple files download related I added a directory and moved all of them there. So the path's changed

PR #399, opened 2021-07-12
GPG-790 Fix bonus pay gap description typo

malinaslevoaca1:
See ticket [GPG-790](https://technologyprogramme.atlassian.net/browse/GPG-790)
"mean" was duplicated because we displayed both "mean" and `Model.Statistic`.
We should use `Model.Statistic` instead of hardcoding "mean".

Tested locally.


Lucas-Fernandes-Nogueira (approved):
Looks great to me! 😁

PR #400, opened 2021-07-12
GPG-783 Fixed invalid reporting year for late submissions bug

malinaslevoaca1:
See ticket [GPG-783](https://technologyprogramme.atlassian.net/browse/GPG-783)

The reporting year was incorrect because the reporting start date (that is used for building that string) was not passed when submitting the late reason and the year for the default value was 1.


Lucas-Fernandes-Nogueira (approved):
Looks goooood! 😀

PR #403, opened 2021-08-09
GPG-792 employers opt out of reporting pay quarters

malinaslevoaca1:
See ticket [GPG-792](https://technologyprogramme.atlassian.net/browse/GPG-792)
Employers can opt out of reporting pay quarters for years with furlough scheme.

Added tests but also tested locally:
* The behavior isn't changed for other reporting years - the pay quarters are required and must add up to 100.
* If not opted out, pay quarters are required and must add up to 100.
If opted out:
* pay quarters cannot be submitted
* pay quarters are marked as "Not applicable" on the check data page
* pay quarters are not displayed on the report page

image
image
image


GenderPayGap.WebUI/Views/AdminOrganisationReturn/ViewReturnDetails.cshtml
```diff
@@ -307,9 +307,9 @@
                                           (thisVersion.FemaleUpperPayBand != previousVersion.FemaleUpperPayBand));
 
                         <td class="govuk-table__cell @(isChanged ? "value-changed-from-previous-return" : null)">
-                            <span style="white-space: nowrap;">@(thisVersion.MaleUpperPayBand.ToString("0"))% male</span>
+                            <span style="white-space: nowrap;">@(thisVersion.MaleUpperPayBand?.ToString("0"))% male</span>
```
Lucas-Fernandes-Nogueira, line 310: Should we not display the table at all if there's no values?

Lucas-Fernandes-Nogueira (approved):
Looks great Malina! Really nice work! I just left a small question but consider this approved 😄.

GenderPayGap.WebUI/Views/AdminOrganisationReturn/ViewReturnDetails.cshtml
malinaslevoaca1: So this table is used for displaying all the versions of a report and it includes all the figures (not only the pay quarters but also the bonuses and the mean, median figures).

Lucas-Fernandes-Nogueira: Will this table sometimes show an empty cell with just the word male in it though?

malinaslevoaca1: Yes, it's the same for the bonus figures that are already optional - they have "%" displayed. We can definitely change that but I don't think it really matters as it's an admin table.

We might want to add a description (eg "Employer opted out of reporting pay quarters") but I will check that with GEO and we can make the changes as part of the other ticket when we update the other copies as well.

Thanks for flagging this!

Lucas-Fernandes-Nogueira: Great, that makes sense. Thanks Malina 😄.

PR #410, opened 2021-09-24
GPG-779: Return all postcodes in all caps

xonde:
Test
Went onto the employer details page and manage org page to make sure the postcode was in all caps. When this is deployed to dev we will be testing the downloaded CSVs

Risk
Since the Postcode automatic getter is still public devs in the future could access it that way, however, this is necessary for setting up entity framework

Docs
N/A

### DIFF
--- a/GenderPayGap.Database/GpgDatabaseContext.OnModelCreating.cs
+++ b/GenderPayGap.Database/GpgDatabaseContext.OnModelCreating.cs
@@ -49,7 +49,7 @@ namespace GenderPayGap.Database
                     entity.Property(e => e.County).HasMaxLength(100);
 
                     entity.Property(e => e.PoBox).HasMaxLength(30);
-                    
+        
                     entity.Property(e => e.PostCode).HasMaxLength(20);
 
                     entity.Property(e => e.Source).HasMaxLength(255);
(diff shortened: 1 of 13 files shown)


GenderPayGap.Database/GpgDatabaseContext.OnModelCreating.cs
```diff
@@ -49,7 +49,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
                     entity.Property(e => e.County).HasMaxLength(100);
 
                     entity.Property(e => e.PoBox).HasMaxLength(30);
-
+
```
malinaslevoaca1, line 52: Could you please revert this? 😁

GenderPayGap.Database/Models/OrganisationAddress.cs
```diff
@@ -56,5 +56,10 @@ public OrganisationAddress()
 
         public virtual ICollection<UserOrganisation> UserOrganisations { get; set; }
 
+        public string GetPostCodeInAllCaps()
+        {
+            return PostCode != null ? PostCode.ToUpper() : PostCode;
```
malinaslevoaca1, line 61: What about `PostCode?.ToUpper()`?

malinaslevoaca1 (commented):
Nice work

GenderPayGap.Database/Models/OrganisationAddress.cs
xonde: ah didn't realise C# had that

GenderPayGap.Database/GpgDatabaseContext.OnModelCreating.cs
xonde: ofc haha

malinaslevoaca1 (approved):

malinaslevoaca1 (dismissed):

malinaslevoaca1 (approved):

PR #418, opened 2021-11-08
GPG-815: Show notice on employer page if retired with submitted reports

xonde:
Mobile:
Mobile - warning centered

Desktop:
Desktop - warning centered


malinaslevoaca1:
I think we should use a different component for adding this. Maybe [inset text](https://design-system.service.gov.uk/components/inset-text/) or [warning text](https://design-system.service.gov.uk/components/warning-text/)?

xonde:
> I think we should use a different component for adding this. Maybe [inset text](https://design-system.service.gov.uk/components/inset-text/) or [warning text](https://design-system.service.gov.uk/components/warning-text/)?

Going off Tor's description on the ticket, I would be inclined to go for the warning text because it's bold, but it gives off the wrong message as it looks like the user has done something wrong. The inset text definitely fits the bill for additional information, but isn't bold (maybe it can be modified)

I'll take screenshots of all the options and share them with Tor on the ticket so she can pick one, thanks for the tip!

PR #419, opened 2021-11-16
GPG-574: IP protect EHRC download endpoint

malinaslevoaca1:
See ticket [GPG-574](https://technologyprogramme.atlassian.net/browse/GPG-574)

We need to IP protect the `download` endpoint that gives EHRC access to some files.
The existing filter was broken due to the load balancer that we have (GoRouter). Considering that we have now the IP deny app I added the whitelisting there, removing the unused (broken) code.
We have a file with all the IPs that should be whitelisted (stored in Azure), configuration is similar to the one we have for the IPs that should be denied.

Tested this on the load test environment.
https://gender-pay-gap-loadtest.london.cloudapps.digital/download?p=App_Data\Downloads\GPG-Organisations_2017-18.csv
Softwire London office IP is whitelisted for now


xonde (approved):
Very DevOps-y, LGTM 🚀

PR #420, opened 2021-11-19
GPG-818 Check report hasn't already been submitted

malinaslevoaca1:
One way to reproduce this bug is to have multiple sessions open at a given time. When the second report is submitted a error message should be displayed informing the user the report has already been submitted.

image

### DIFF
--- a/GenderPayGap.WebUI/Controllers/SubmitController.CheckData.cs
+++ b/GenderPayGap.WebUI/Controllers/SubmitController.CheckData.cs
@@ -9,6 +9,7 @@ using GenderPayGap.Database;
 using GenderPayGap.WebUI.BusinessLogic.Classes;
 using GenderPayGap.WebUI.BusinessLogic.Models.Submit;
 using GenderPayGap.WebUI.Classes;
+using GenderPayGap.WebUI.ErrorHandling;
 using GenderPayGap.WebUI.Models.Submit;
 using GenderPayGap.WebUI.Services;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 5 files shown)


GenderPayGap.WebUI/Controllers/SubmitController.CheckData.cs
```diff
@@ -174,11 +175,20 @@ public async Task<IActionResult> CheckData(ReturnViewModel postedReturnViewModel
             }
 
             Return postedReturn = submissionService.CreateDraftSubmissionFromViewModel(postedReturnViewModel);
+            Return databaseReturn = submissionService.GetReturnFromDatabase(postedReturnViewModel.OrganisationId, postedReturnViewModel.AccountingDate.Year);
 
             SubmissionChangeSummary changeSummary = null;
-            Return databaseReturn = submissionService.GetSubmissionById(postedReturnViewModel.ReturnId);
             if (databaseReturn != null)
             {
+                if (databaseReturn.ReturnId != postedReturnViewModel.ReturnId)
```
xonde, line 183: What does this check for?

xonde (approved):
All looks good! Just a question about one of the checks

GenderPayGap.WebUI/Controllers/SubmitController.CheckData.cs
malinaslevoaca1: this checks if the report has already been submitted (not this report - different id). Basically you can edit an old report (it will be marked as retired later) but can't have 2 submitted  reports at the same time.

xonde (approved):
NOICE! 🚀

PR #422, opened 2021-12-06
GPG-823: Trim user input

malinaslevoaca1:


xonde (approved):
Quite a lot to review here, but it looks all good 👀

PR #423, opened 2021-12-08
GPG-578 change password improvements

malinaslevoaca1:
https://technologyprogramme.atlassian.net/browse/GPG-578
D: None
R: None
T: tested locally as well

### DIFF
--- a/GenderPayGap.WebUI/Controllers/Account/ChangePasswordController.cs
+++ b/GenderPayGap.WebUI/Controllers/Account/ChangePasswordController.cs
@@ -66,6 +66,7 @@ namespace GenderPayGap.WebUI.Controllers.Account
             {
                 currentUser.ResetAttempts = 0;
                 dataRepository.SaveChanges();
+                StatusMessageHelper.SetStatusMessage(Response, "You have been logged out for security reasons", Url.Action("LoggedOut", "Login"));
                 return LoginHelper.Logout(HttpContext, RedirectToAction("LoggedOut", "Login"));
             }
 
@@ -89,7 +90,7 @@ namespace GenderPayGap.WebUI.Controllers.Account
         public void ValidatePasswords(ChangePasswordViewModel viewModel, User currentUser)
         {
             // Check if current password is correct
-            bool isValidPassword = userRepository.CheckPassword(currentUser, viewModel.CurrentPassword, false);
+            bool isValidPassword = userRepository.CheckPassword(currentUser, viewModel.CurrentPassword, true);
             if (!isValidPassword)
             {
                 viewModel.AddErrorFor(m => m.CurrentPassword, "Could not verify your current password.");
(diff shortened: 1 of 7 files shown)


GenderPayGap.WebUI/Controllers/Account/ChangePasswordController.cs
```diff
@@ -61,6 +62,13 @@ public IActionResult ChangePasswordPost(ChangePasswordViewModel viewModel)
             
             // Check that passwords are valid
             ValidatePasswords(viewModel, currentUser);
+            if (currentUser.ResetAttempts == Global.MaxLoginAttempts)
+            {
+                currentUser.ResetAttempts = 0;
+                dataRepository.SaveChanges();
+                StatusMessageHelper.SetStatusMessage(Response, "You have been logged out for security reasons", Url.Action("LoggedOut", "Login"));
+                return LoginHelper.Logout(HttpContext, RedirectToAction("LoggedOut", "Login"));
```
xonde, line 70: can the user attempt to carry on changing their password after they are logged out as we've set their reset attempts back to 0?

GenderPayGap.WebUI/Controllers/Account/ChangePasswordController.cs
```diff
@@ -61,6 +62,13 @@ public IActionResult ChangePasswordPost(ChangePasswordViewModel viewModel)
             
             // Check that passwords are valid
             ValidatePasswords(viewModel, currentUser);
+            if (currentUser.ResetAttempts == Global.MaxLoginAttempts)
```
xonde, line 65: it might be worth setting a new global variable for Max reset attempts as well, what do you think?

xonde (commented):
Just a couple of questions about the max reset attemps everything else looks all good to me

GenderPayGap.WebUI/Controllers/Account/ChangePasswordController.cs
malinaslevoaca1: yes but their account is locked for 30 minutes - same behavior when introducing the wrong credentials at login

GenderPayGap.WebUI/Controllers/Account/ChangePasswordController.cs
malinaslevoaca1: I find them somehow related (i.e. introducing wrong credentials) so that's why I used the same variable - however, if using it I should rename it.

Do you think it's worth adding a new variable for this?

xonde: Yeah, a rename would be just as good as well, just something to show that this variable is meant to be used here, maybe something like `Global.MaxAuthAttemps`? (feel free to not use that suggestion 😅)

GenderPayGap.WebUI/Controllers/Account/ChangePasswordController.cs
xonde: Ah brill, wasn't sure where the lock out logic was

xonde (approved):
Nice one 🚀

PR #424, opened 2021-12-08
GPG-810 Soft delete organisation if registration request is rejected

malinaslevoaca1:
See [this comment](https://technologyprogramme.atlassian.net/browse/GPG-810?focusedCommentId=75866)

### DIFF
--- a/GenderPayGap.WebUI/Controllers/RegisterController.Request.cs
+++ b/GenderPayGap.WebUI/Controllers/RegisterController.Request.cs
@@ -612,7 +612,7 @@ namespace GenderPayGap.WebUI.Controllers
         [HttpPost("confirm-cancellation")]
         public async Task<IActionResult> ConfirmCancellation(OrganisationViewModel model, string command)
         {
-            //Ensure user has completed the registration process
+            // Ensure user has completed the registration process
             User currentUser;
             IActionResult checkResult = CheckUserRegisteredOk(out currentUser);
             if (checkResult != null)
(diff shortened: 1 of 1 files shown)


GenderPayGap.WebUI/Controllers/RegisterController.Request.cs
```diff
@@ -686,6 +686,12 @@ public async Task<IActionResult> ConfirmCancellation(OrganisationViewModel model
                     $"Unused organisation {userOrg.OrganisationId}:'{userOrg.Organisation.OrganisationName}' deleted by {(OriginalUser == null ? currentUser.EmailAddress : OriginalUser.EmailAddress)} when declining manual registration for {userOrg.User.EmailAddress}");
                 DataRepository.Delete(userOrg.Organisation);
             }
+            if (userOrg.Organisation != null
+                && userOrg.Organisation.UserOrganisations.Count == 1
+                && userOrg.Organisation.Status == OrganisationStatuses.Pending)
+            {
+                userOrg.Organisation.SetStatus(OrganisationStatuses.Deleted, details: "Manually Rejected");
```
xonde, line 693: Random question, but why do we have the organisation in the database if they haven't been registered already? Actually what exactly are they registering for?

xonde (commented):
Just a question about the registration process, but code looks good to me

GenderPayGap.WebUI/Controllers/RegisterController.Request.cs
malinaslevoaca1: when adding an organisation to their account a user (the user does that so that they can report for that org) can either:
* select an existing organisation (that is already in our db OR is fetched from Companies House) OR
* add a new organisation - the organisation details being provided by the user
In the second case, the organisation is added to the database but its status is Pending. All the other organisations are Active

All the registrations need to be approved by an admin - if the registration is rejected the Pending organisations should be deleted so that they can't be used by other users

xonde: Is the soft delete for posterity's sake or because the org can be re-activated at a later date?

malinaslevoaca1: for posterity's sake - I mean it can be reactivated as well but I don't think this will happen

xonde: Ah cool, sorry I thought I had approved this!

xonde (approved):

PR #425, opened 2022-01-10
GPG-805: Remove draft if no changes

malinaslevoaca1:
see ticket: https://technologyprogramme.atlassian.net/browse/GPG-805

The draft must be removed from the database if there are no changes saved.

### DIFF
--- a/GenderPayGap.WebUI/BusinessLogic/Services/DraftFileBusinessLogic.cs
+++ b/GenderPayGap.WebUI/BusinessLogic/Services/DraftFileBusinessLogic.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using GenderPayGap.Core;
 using GenderPayGap.Core.Interfaces;
+using GenderPayGap.Database;
 using GenderPayGap.Database.Models;
 using GenderPayGap.Extensions;
 using GenderPayGap.WebUI.BusinessLogic.Classes;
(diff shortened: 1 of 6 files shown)


GenderPayGap.WebUI/BusinessLogic/Services/DraftFileBusinessLogic.cs
```diff
@@ -97,11 +98,13 @@ public void KeepDraftFileLockedToUser(Draft draftExpectedToBeLocked, long userId
             SetDraftAccessInformation(userIdRequestingLock, draftExpectedToBeLocked);
         }
 
-        public void DiscardDraft(Draft draftToDiscard)
+        public void DiscardDraft(Draft draftToDiscard, Return submittedReturn)
         {
             List<DraftReturn> matchingReturns = dataRepository.GetAll<DraftReturn>()
                 .Where(dr => dr.OrganisationId == draftToDiscard.OrganisationId)
                 .Where(dr => dr.SnapshotYear == draftToDiscard.SnapshotYear)
+                .AsEnumerable()
```
xonde, line 106: Why are we calling this?

xonde (commented):
Look all good, just a question about one line

GenderPayGap.WebUI/BusinessLogic/Services/DraftFileBusinessLogic.cs
malinaslevoaca1: EF attempts to evaluate the query on the server where possible. However, on the next line we are calling some external functions that can't be translated into SQL automatically. Therefore, the query needs to be evaluated on the client side here - the data needs to be downloaded into the memory and further parsed (this happens when we call `AsEnumerable()`). Otherwise, an error is thrown.

xonde: Ah, I see, thanks for the explanation!

xonde (approved):
LGTM 🚀

PR #427, opened 2022-01-13
GPG-826: Trust AWS Cloudfront IP ranges

malinaslevoaca1:
We found out the nginx app didn't work in Prod when checking if the EHRC endpoints were IP protected. Turned out the app was not configured correctly (as in it didn't work with a cdn routed domain)

You can see the gov.uk PaaS support reply here: https://technologyprogramme.atlassian.net/browse/GPG-826

Tested the configuration on the load test environment


xonde (approved):
Nice stuff!

PR #431, opened 2022-02-09
GPG-830: Bump npm modules

malinaslevoaca1:
See ticket [GPG-830](https://technologyprogramme.atlassian.net/browse/GPG-830)

There were some breaking changes so we need to make some code changes as well.
All the breaking changes are listed [here](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md) and [here](https://github.com/sass/node-sass/releases).
Components affected:
* Accordion - the design has changed. Page: Actions to close the gap and Manual SIC codes
* Header
* Cookie banner
* Hints - Page: Admin search page
* Skip link
* Error messages
* Conditional check boxes - Page: Feedback page
* Characters count - Page: Feedback page

The designs should look the same apart from the accordions that use[ the new design](https://design-system.service.gov.uk/components/accordion/)
**Testing:**
Checked different journeys and the service looks fine. 
**Risk:**
I didn't check all the pages. However, before releasing this all the journeys will be retested to ensure the app works as expected.
**Documentation:** None

### DIFF
--- a/GenderPayGap.WebUI/Views/AddOrganisationManualConfirm/ManualConfirm.cshtml
+++ b/GenderPayGap.WebUI/Views/AddOrganisationManualConfirm/ManualConfirm.cshtml
@@ -61,10 +61,10 @@
                     <td class="govuk-table__cell">
                         @if (Model.HasErrorFor(m => m.OrganisationName))
                         {
-                            <span class="govuk-error-message" id="GovUk_@(nameof(Model.OrganisationName))-error">
+                            <p class="govuk-error-message" id="GovUk_@(nameof(Model.OrganisationName))-error">
                                 <span class="govuk-visually-hidden">Error</span>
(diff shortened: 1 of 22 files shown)


GenderPayGap.WebUI/Views/AddOrganisationManualConfirm/ManualConfirm.cshtml
```diff
@@ -61,10 +61,10 @@
                     <td class="govuk-table__cell">
                         @if (Model.HasErrorFor(m => m.OrganisationName))
                         {
-                            <span class="govuk-error-message" id="GovUk_@(nameof(Model.OrganisationName))-error">
+                            <p class="govuk-error-message" id="GovUk_@(nameof(Model.OrganisationName))-error">
```
xonde, line 64: Why are these being changed from spans to p?

PR #433, opened 2022-02-25
Ensure "Maintenance Mode" doesn't break Health Check

jamesgriff:
Hi Malina,
I'm not sure if this works (I haven't been able to run it on my laptop - I don't have Postgres installed!), but this is the code I think we'd need to allow Maintenance Mode to work.


malinaslevoaca1 (approved):
Looks good to me - tested on the load test environment: https://gender-pay-gap-loadtest.london.cloudapps.digital/

PR #437, opened 2022-03-15
GPG-842: Get return by reporting year

malinaslevoaca1:
See details: https://technologyprogramme.atlassian.net/browse/GPG-842

When the sector of an organisation is changed all the accounting dates in the database are updated (for all the returns and scopes). This issue is caused by the following: the sector of an organisation was updated without updating the accounting date as well.

When fetching the return from the database we should compare the years of the accounting dates.

### DIFF
--- a/GenderPayGap.Database/Models/Extensions/Organisation.cs
+++ b/GenderPayGap.Database/Models/Extensions/Organisation.cs
@@ -324,17 +324,17 @@ namespace GenderPayGap.Database
                 .FirstOrDefault();
         }
 
-        //Returns the latest return for the specified accounting year or the latest ever if no accounting year is 
+        // Returns the latest return for the specified accounting year or the latest ever if no accounting year is 
         public Return GetReturn(int year = 0)
         {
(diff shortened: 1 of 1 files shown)


GenderPayGap.Database/Models/Extensions/Organisation.cs
```diff
@@ -324,17 +324,17 @@ public Return GetLatestReturn()
...
         }
 
-        //Returns the latest return for the specified accounting year or the latest ever if no accounting year is 
+        // Returns the latest return for the specified accounting year or the latest ever if no accounting year is 
         public Return GetReturn(int year = 0)
         {
             DateTime accountingStartDate = SectorType.GetAccountingStartDate(year);
             return Returns
-                .Where(r => r.Status == ReturnStatuses.Submitted && r.AccountingDate == accountingStartDate)
+                .Where(r => r.Status == ReturnStatuses.Submitted && r.AccountingDate.Year == accountingStartDate.Year)
```
xonde, line 332: Looks all good, just wanted to check the issue was as I cannot see the JIRA page yet 😭

xonde (approved):

PR #438, opened 2022-03-15
GPG-844: Check if property has no errors before adding others

malinaslevoaca1:
see the bug details [here](https://technologyprogramme.atlassian.net/browse/GPG-844)

Tested locally

### DIFF
--- a/GenderPayGap.WebUI/Helpers/ReportFiguresHelper.cs
+++ b/GenderPayGap.WebUI/Helpers/ReportFiguresHelper.cs
@@ -300,30 +300,20 @@ namespace GenderPayGap.WebUI.Helpers
             HttpRequest request, 
             Expression<Func<ReportFiguresViewModel, decimal?>> propertyLambdaExpression)
         {
-            ValidateParameterIfNoErrors(
-                viewModel,
-                propertyLambdaExpression,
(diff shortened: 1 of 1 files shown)


GenderPayGap.WebUI/Helpers/ReportFiguresHelper.cs
```diff
@@ -298,11 +298,32 @@ private static void ValidateHourlyPayFigures(ReportFiguresViewModel viewModel, H
...
+            ValidateParameterIfNoErrors(
+                viewModel,
+                propertyLambdaExpression,
+                () => viewModel.AddErrorFor(propertyLambdaExpression, errorMessage));
+        }
+
+        private static void ValidateParameterIfNoErrors(
+            ReportFiguresViewModel viewModel, 
+            Expression<Func<ReportFiguresViewModel, decimal?>> propertyLambdaExpression,
+            Action callback)
```
xonde, line 322: We could give this a more descriptive name like `validatorFunction` or `validatorCallback`

xonde: Also is there a reason we use `Action` and not `Func`?

malinaslevoaca1: agreed 👍

PR #441, opened 2022-03-17
GPG-843: Get OptOut bool when creating DraftReturn from SubmittedReturn

xonde:


GenderPayGap.WebUI/Services/DraftReturnService.cs
```diff
@@ -61,6 +61,8 @@ private DraftReturn CreateDraftReturnBasedOnSubmittedReturn(Return submittedRetu
             {
                 OrganisationId = submittedReturn.OrganisationId,
                 SnapshotYear = submittedReturn.AccountingDate.Year,
+                
+                OptedOutOfReportingPayQuarters = submittedReturn.OptedOutOfReportingPayQuarters,
```
malinaslevoaca1, line 65: why didn't this issue happen when editing the figures first?

xonde: We explicitly save the opt out of reporting boolean when we edit the figures

malinaslevoaca1 (approved):

PR #443, opened 2022-03-17
GPG-845: Users are now redirected to the appropriate page when reporting

xonde:
Figures -> Person responsible (if private) -> Org size -> Link to GPG info -> Report overview


malinaslevoaca1 (commented):
this should happen only when the report is edited for the first time, is that the case here?

malinaslevoaca1:
@xonde I will merge this as it is. will make any other required changes later

PR #447, opened 2022-03-29
GPG-857: Update email field validation to match GDS pattern

laurasaunders1:


malinaslevoaca1 (approved):
Looks great 😄 

I'll mark the PR as draft as we shouldn't merge anything until the deadline has passed (just in case there are some fixes we need to release around that time)

PR #448, opened 2022-04-13
GPG-856: User registration

laurasaunders1:


GenderPayGap.WebUI/Controllers/Account/AccountCreationController.cs
```diff
@@ -122,16 +122,27 @@ public IActionResult CreateUserAccountPost(CreateUserAccountViewModel viewModel)
...
 
             if (viewModel.HasAnyErrors())
             {
                 return View("CreateUserAccount", viewModel);
             }
 
-            User newUser = CreateNewUser(viewModel);
-            dataRepository.Insert(newUser);
+            // Check for retired account with matching email address
+            User retiredUser = userRepository.FindByEmail(viewModel.EmailAddress, UserStatuses.Retired);
```
laurasaunders1, line 133: This looks up the email address and does FirstOrDefault() so will not error if there are multiple matches, although does mean that additional matches might not be removed/ reactivated. The clean up job should resolve this issue though, as it will remove duplicated retired accounts.

GenderPayGap.WebUI/Controllers/Account/AccountCreationController.cs
```diff
@@ -193,6 +204,32 @@ private User CreateNewUser(CreateUserAccountViewModel viewModel)
...
+            user.HashingAlgorithm = HashingAlgorithm.PBKDF2;
+
+            user.EmailVerifySendDate = null;
+            user.EmailVerifiedDate = null;
+            user.EmailVerifyHash = null;
+            user.SetStatus(UserStatuses.New, user,  "Retired user account has been reactivated");
+            user.Status = UserStatuses.New;
+
+            return user;
+        }
```
malinaslevoaca1, line 231: these 2 methods look very similar. I suggest we refactor them to remove the duplicated code, what do you think?

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseRetiredDuplicateUsersJob.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            List<User> users = dataRepository.GetAll<User>().ToList();
+
+            var usersWithMultipleAccounts = users
+                .Where(u => u.HasBeenAnonymised == false)
+                .GroupBy(u => u.EmailAddress).Where(e => e.Count() > 1)
+                .SelectMany(g => g)
+                .ToList();
+
+            foreach (var user in usersWithMultipleAccounts.Where(u => u.Status == UserStatuses.Retired))
```
malinaslevoaca1, line 40: we also anonymise the users that have more retired accounts (no active one). I'm wondering if we need to keep the data for those users

malinaslevoaca1 (commented):
Nice work 👍

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseRetiredDuplicateUsersJob.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            JobHelpers.RunAndLogJob(AnonymiseRetiredDuplicateUsersAction, nameof(AnonymiseRetiredDuplicateUsers));
+        }
+
+        private void AnonymiseRetiredDuplicateUsersAction()
+        {
+            List<User> users = dataRepository.GetAll<User>().ToList();
+
+            var usersWithMultipleAccounts = users
+                .Where(u => u.HasBeenAnonymised == false)
```
malinaslevoaca1, line 35: what about `.Where(u => !u.HasBeenAnonymised)`

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseRetiredDuplicateUsersJob.cs
laurasaunders1: Do you mean something like, if all retired anonymise all but the most recently created account?

malinaslevoaca1: yes, just in case that data is needed for something (I'm not sure if it is). What do you think?

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseRetiredDuplicateUsersJob.cs
laurasaunders1: Well this is embarrassing! :) Good catch

GenderPayGap.WebUI/Controllers/Account/AccountCreationController.cs
laurasaunders1: Yeah, that seems sensible!

GenderPayGap.WebUI/BackgroundJobs/ScheduledJobs/AnonymiseRetiredDuplicateUsersJob.cs
laurasaunders1: Yeah, I wasn't 100% sure tbh, I did change my mind a few times. I feel like the data is not needed really, but this does feel more complete and considered, removing only the minimum amount of data

laurasaunders1:
Clean up job skips the first user in the list, this orders first by status unknown > new > active > retired and then by created date desc. This means that newest user will not be anonymised. (We additionally check so that only retired users are anonymised, although there should not be more than 1 non-retired user account in theory)

malinaslevoaca1 (approved):

PR #453, opened 2022-07-22
GPG-482-download-file-sizes: New structure for list generation

Airk0n:


GenderPayGap.WebUI/Views/Viewing/Download/Download.cshtml
```diff
@@ -1,9 +1,14 @@
...
 
-@{
+@{
     ViewBag.Title = "Download gender pay gap figures - Gender pay gap service";
-    Layout = "~/Views/GovUkFrontend/GovUkFrontendLayout.cshtml";
+    Layout = "~/Views/GovUkFrontend/GovUkFrontendLayout.cshtml";
+}
+
+@{
+    Dictionary<int, int> YearAndSize = new Dictionary<int, int>();
```
davidm-m, line 11: This shouldn't need to be in its own separate code block

davidm-m (changes requested):
Generally happy but confused why package-lock.json has been deleted. If that wasn't deliberate it should be restored.
I'd also be really grateful if you could add a screenshot of the download page so I can see your changes in action 🙂

Airk0n:
> 
image
Image of file sizes added.

As for the package-lock.json I'm a little confused, firstly it's showing as nested within package.json in VS: 
image
But in terms of file directory it's actually in the same directory. Also it looks like the changes haven't been made on my end (the file is still there) so it's not showing that I need to commit the change.

GenderPayGap.WebUI/Views/Viewing/Download/Download.cshtml
```diff
@@ -30,15 +32,18 @@
...
-                </li>
+            @{
+                List <(int,string)> reportingYearsWithFileSizes = ReportingYearsHelper.GetReportingYearWithCSVFileSize();
+                foreach (var item in reportingYearsWithFileSizes)
+                {
+                    <li>
+                        <a href="@Url.Action("DownloadData", "Viewing", item.Item1)"
+                           class="govuk-link">
+                            <span class="govuk-visually-hidden">Download gender pay gap data for</span>
+                            Reporting year @(ReportingYearsHelper.FormatYearAsReportingPeriod(item.Item1)) (@item.Item2 .CSV file)
```
davidm-m, line 43: Didn't notice this before - previously this didn't have a dot so it was just "CSV file". Very minor but it would be good to not change this unnecessarily, could we get rid of the dot again?

GenderPayGap.WebUI/Views/Viewing/Download/Download.cshtml
```diff
@@ -1,9 +1,11 @@
 ﻿@using GenderPayGap.Core.Helpers
 @using GenderPayGap.WebUI.Models.Shared
+@using System.IO
 
-@{
+@{
     ViewBag.Title = "Download gender pay gap figures - Gender pay gap service";
-    Layout = "~/Views/GovUkFrontend/GovUkFrontendLayout.cshtml";
+    Layout = "~/Views/GovUkFrontend/GovUkFrontendLayout.cshtml";
+    Dictionary<int, int> YearAndSize = new Dictionary<int, int>();
```
davidm-m, line 8: I don't think this dictionary is getting used anywhere any more.

davidm-m (approved):
Looks good!

GenderPayGap.Core/Helpers/ReportingYearsHelper.cs
```diff
@@ -1,77 +1,104 @@
...
+            List<int> reportingYears = GetReportingYears().OrderByDescending(y => y).ToList<int>();
+            List<(int, int)> yearsWithFileSizes = new List<(int, int)>()
+            {
+                (reportingYears[0], 55),
+                (reportingYears[1], 4032),
+                (reportingYears[2], 4129),
+                (reportingYears[3], 2798),
+                (reportingYears[4], 4266),
+                (reportingYears[5], 4211),
+            };
```
jamesgriff, line 32: Errrrm, it looks awfully like the file sizes here are hard-coded! 😱
Over time, as more organisations report their GPG, these file sizes will change.

Also, this method will only ever return 6 items.
Currently, there are 6 reporting years, but once we move into the next reporting year, there will be 7.
And it's wrong to not show all available files.

If we're going to fix a minor usability issue like this (adding file sizes), it's extra important not to break any existing functionality.

CC: @davidm-m

PR #454, opened 2022-08-01
GPG-484: "Error:" added to title based on if the model has any errors.

Airk0n:
GPG a

"Error:" now added to page title based on `Model.HasAnyErrors();`


davidm-m (approved):
Looks good!

PR #455, opened 2022-08-03
GPG-473-tabs-on-compare: aria-selected and role added

Airk0n:
tbas
selected
This ticket fixes an issue where the aria-selected labels were either not showing or not responding to the currently selected year.
This also fixes an issue where the role was not set to "tab" which caused screen readers to display each of the tab links as "links" they now show as selected and unselected tabs (as seen in the attached screenshots).


davidm-m (approved):
Nice stuff 👍

PR #456, opened 2022-08-09
GPG-472: Bar charts & keys have outline, screen reader ignore keys.

Airk0n:
### **Changes made :**
- Transparent border added to bar charts(both horizontal and vertical) this displays their outline on high contrast mode only.
- Transparent border added to keys "Women, Men", this displays their outline on high contrast mode only.
- Bar chart keys "Women, Men" hidden from screen-reader : they keys are irrelevant to screen readers as the data given by the bar charts is already given in text form without a colour key.

### **High Contrast mode with outlines visible:**
High contrast borderes added

### **Screen reader ignoring bar graphs and keys and reading out the visually hidden data.**
Screen reader

### DIFF
--- a/GenderPayGap.WebUI/Content/site.charts.css
+++ b/GenderPayGap.WebUI/Content/site.charts.css
@@ -103,8 +103,7 @@ govuk-chart-vert__head {
     width: 4em;
     height: 100%;
     bottom: 0;
-    outline: 2px solid;
-    outline-color: transparent;
+    outline: 2px solid transparent;
 }
 
 .govuk-chart-vert__bar--women, .govuk-chart-vert__caption--women {
--- a/GenderPayGap.WebUI/Content/site.charts.css
+++ b/GenderPayGap.WebUI/Content/site.charts.css
@@ -150,8 +150,7 @@ govuk-chart-vert__head {
 
 .govuk-chart-horz__bar {
     height: 2em;
-    outline: 2px solid;
-    outline-color: transparent;
+    outline: 2px solid transparent;
 }
 
 .govuk-chart-horz__bar--men {
--- a/GenderPayGap.WebUI/Content/site.charts.css
+++ b/GenderPayGap.WebUI/Content/site.charts.css
@@ -18,10 +18,12 @@
 
 .govuk-legend__caption--women:before {
 	background-color: #F46A25;
+    outline: 2px solid transparent;
 }
 
 .govuk-legend__caption--men:before {
-	background-color: #919397;
+    background-color: #919397;
+    outline: 2px solid transparent;
 }
 
 .govuk-chart-vert {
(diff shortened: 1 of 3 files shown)


GenderPayGap.WebUI/Content/site.charts.css
```diff
@@ -101,6 +103,8 @@ govuk-chart-vert__head {
     width: 4em;
     height: 100%;
     bottom: 0;
+    outline: 2px solid;
+    outline-color: transparent;
```
davidm-m, line 107: Any particular reason why in some we do `2px solid transparent` but here we do `2px solid; outline-color: transparent`?

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Quartiles/QuartilesBarChart.cshtml
```diff
@@ -14,10 +14,10 @@
 
     <div class="govuk-chart-horz__foot">
         <div class="govuk-chart-horz__text--left">
-            <span class="bold-small">@(Model.FemaleFormatted)% <span class="visually-hidden"> of the @Model.Title.ToLower() are women</span></span>
+            <span class="bold-small">@(Model.FemaleFormatted)% <span class="visually-hidden"> &nbsp of the @Model.Title.ToLower() are women</span></span>
```
davidm-m, line 17: Why are we adding non-breaking spaces here? I also think this isn't right, they should have a `;` at the end.

davidm-m (changes requested):
Got a couple of questions about this before I approve it.

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Quartiles/QuartilesBarChart.cshtml
Airk0n: image
With and without the non-breaking space, interestingly it doesn't care about the ; but I will add it.

GenderPayGap.WebUI/Content/site.charts.css
```diff
@@ -147,6 +150,8 @@ govuk-chart-vert__head {
 
 .govuk-chart-horz__bar {
     height: 2em;
+    outline: 2px solid;
+    outline-color: transparent;
```
davidm-m, line 154: This one's still separated into multiple lines

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Quartiles/QuartilesBarChart.cshtml
davidm-m: Is that two spaces in the one with the nbsp? Probably doesn't matter since it will be read out. Have you tried actually listening to it?

davidm-m (changes requested):

davidm-m (approved):
Good stuff

PR #459, opened 2022-08-09
GPG-475: Comparison list changes changed to role="alert".

Airk0n:
**Change made:**
 - The role of the message that shows how many companies you have in your comparison bin is changed from "status" to "alert" which causes the screen reader to read out how many we are comparing and gives the user immediate feedback.
 
 
Comparison bin on screenreader


davidm-m (approved):
Looks good to me!

PR #461, opened 2022-08-15
GPG-898: removed js setting tabIndex attribute and cleanup.

Airk0n:
We are now able to tab to the first year on the employer comparison page - 
image

Previously the first year's tabIndex was being set to -1 and this caused it to be inaccessible via the tab key.


davidm-m:
Since we're not changing our package.json at all I don't think we want to commit a change to package-lock.json.

davidm-m (changes requested):
This shouldn't be updating `package-lock.json` at all as were not trying to change our installed packages, and the changes to `package-lock.json` look very questionable too.

GenderPayGap.WebUI/wwwroot/scripts/gov_uk_front_end.js
```diff
@@ -2647,30 +2647,22 @@
...
 		var panelId = this.getHref($tab).slice(1);
-		$tab.setAttribute('id', 'tab_' + panelId);
-		$tab.setAttribute('role', 'tab');
 		$tab.setAttribute('aria-controls', panelId);
 		$tab.setAttribute('aria-selected', 'false');
-		$tab.setAttribute('tabindex', '-1');
 
 		// set panel attributes
 		var $panel = this.getPanel($tab);
-		$panel.setAttribute('role', 'tabpanel');
```
davidm-m, line 2658: Remind me why we're removing this one? I _think_ it's because it's already being set anyway but I just want to be sure.

davidm-m (approved):
I think it's fine, got one question but shouldn't need any code changes.

GenderPayGap.WebUI/wwwroot/scripts/gov_uk_front_end.js
Airk0n: Yes, the JS isn't working properly (which is in another ticket) so we removed this attribute so it's clear that's it's only being set in the views with C#/Html.

PR #462, opened 2022-08-16
GPG-725: Fix back button logic when viewing comparison table

CodrinCotarlanSoftwire:


GenderPayGap.WebUI/Views/Compare/CompareTableRows.cshtml
```diff
@@ -16,10 +16,22 @@
...
+                        string id = compareReport.EncOrganisationId;
+                    }
+                    <a id="remove-row-{@id}" href="@removeRowLink">Remove <span class="visually-hidden">@compareReport.OrganisationName from comparison</span></a>
+                    <script>
+                        document.getElementById("remove-row-{@id}").addEventListener("click", (e) => {
+                            e.stopPropagation()
+                            e.preventDefault()
+                            window.location.replace("@removeRowLink");
+                        })
+                    </script>
```
CodrinCotarlanSoftwire, line 32: Problem: Clicking 'delete' makes a post request which redirects to the same page, so clicking 'back' would just go back to the previous page which, in this case, would've been the same.

Solution: Do not add the current page to the browser history.
Note: I have to add an href attribute to `<a>` to keep the link appearance, but in practice it has 0 functionality since `window.location.replace` does all the work...

davidm-m (approved):
I wondered whether we'd want to keep this in an actual javascript file rather than have it as an inline script, but it looks like throughout the project we regularly use one-off inline scripts, so it should be fine.

Looks good, thanks!

PR #464, opened 2022-08-22
Gpg 868 migrate database

CodrinCotarlanSoftwire:


terraform/main.tf
```diff
@@ -1,15 +1,62 @@
...
-  versioning = {
-    enabled = true
-  }
+resource "aws_security_group_rule" "postgres_in" {
+  security_group_id = aws_security_group.allow_postgres_connection.id
+  type              = "ingress"
+  from_port         = local.postgres_config.port
+  to_port           = local.postgres_config.port
+  protocol          = "tcp"
+  cidr_blocks       = ["0.0.0.0/0"]
```
davidm-m, line 32: Do we really want to allow ingress and egress from any ip address? Is there a way to restrict this to just connections from within our vpc for instance? I don't think we need any external applications to directly connect to the database.

terraform/prod-app.tfvars
```diff
@@ -0,0 +1,10 @@
+postgres_config = {
+  instance_class = "db.t3.small"
+  identifier     = "gpg-prod-db"
+  db_name        = "gpgProdDb"
+  username       = "postgres"
+  password       = "postgres"
```
davidm-m, line 6: For my peace of mind, is there a future ticket to change this so we're not storing our database credentials in plain text, in the code? If so it would be nice to have a todo comment here referencing it.

davidm-m (commented):
Mostly looks good, I've got a couple of questions I'd like answered before it's merged though.

terraform/main.tf
```diff
@@ -1,15 +1,62 @@
...
 
-module "example" {
-  source = "terraform-aws-modules/s3-bucket/aws"
-  
-  bucket = "gpg-bucket-${terraform.workspace}"
-  acl    = "private"
+resource "aws_security_group" "allow_postgres_connection" {
+  name        = join("_", ["allow_postgres_connection", var.env])
+  description = "Allow Postgres DB traffic"
+  vpc_id      = "vpc-0e9491851e4208470"
```
LadunOmideyiSoftwire, line 23: vpc_id can just be a variable e.g. var.vpc_id as this will be defined in the vpc resource

So can subnet (private_subnets) but I cannot see this defined anywhere. Where are we specifying the rds subnet?

terraform/main.tf
```diff
@@ -1,15 +1,62 @@
...
+// declarations for the resources to be created 
 // this does not include the state S3 bucket as this must be created in order for terraform to init
 
+locals {
+  postgres_config = defaults(var.postgres_config, {
+    allocated_storage           = 100
+    engine                      = "postgres"
+    engine_version              = "14"
+    port                        = 5432
+    backup_retention_period     = 5
```
LadunOmideyiSoftwire, line 10: We should make longer. Maybe 15 - 30 days?

terraform/main.tf
```diff
@@ -1,15 +1,62 @@
...
+  from_port         = 0
+  to_port           = 0
+  protocol          = "-1"
+  cidr_blocks       = ["0.0.0.0/0"]
+}
+
+resource "aws_db_instance" "gpg-dev-db" {
+  allocated_storage           = local.postgres_config.allocated_storage
+  engine                      = local.postgres_config.engine
+  engine_version              = local.postgres_config.engine_version
```
LadunOmideyiSoftwire, line 47: Does it matter that this is a newer version than used in the codebase?

LadunOmideyiSoftwire (changes requested):

terraform/prod-app.tfvars
CodrinCotarlanSoftwire: Yes, I've moved them as environment variables in devops.

terraform/main.tf
```diff
@@ -1,15 +1,62 @@
...
+  password                    = local.postgres_config.password
+  port                        = local.postgres_config.port
+  backup_retention_period     = local.postgres_config.backup_retention_period
+  backup_window               = local.postgres_config.backup_window
+  vpc_security_group_ids      = [aws_security_group.allow_postgres_connection.id]
+  storage_encrypted           = local.postgres_config.storage_encrypted
+  publicly_accessible         = local.postgres_config.publicly_accessible
+  allow_major_version_upgrade = local.postgres_config.allow_major_version_upgrade
+  multi_az                    = local.postgres_config.multi_az
+  skip_final_snapshot         = local.postgres_config.skip_final_snapshot
```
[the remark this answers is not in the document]
CodrinCotarlanSoftwire: I don't think there's any reason. I couldn't find whether the current configuration skips it or not. There's no harm in _not_ skipping it. I'll Add it.

terraform/main.tf
CodrinCotarlanSoftwire: Was not sure about the number. I know gov paas performs weekly backups. I'll put 15 in there then.

terraform/main.tf
CodrinCotarlanSoftwire: It should make no difference. Locally I installed the latest version by default and it works without issue, so this should too.

terraform/variables.tf
```diff
@@ -24,8 +24,8 @@ variable "postgres_config" {
     instance_class              = string
     identifier                  = string
     db_name                     = string
-    username                    = string
-    password                    = string
+    username                    = optional(string)
+    password                    = optional(string)
     port                        = optional(number)
```
davidm-m, line 29: What happens if these aren't provided? I assume the setup will fail, in which case they probably shouldn't be optional?

CodrinCotarlanSoftwire:
Continuing from our work with @LadunOmideyiSoftwire, I have, in essence:

- added the VPC module (instead of creating the resources)
- Removed the `defaults` experimental module
- split each variable in the postgres object into it's own (each with its own description and sometimes default)

terraform/.terraform.lock.hcl
```diff
@@ -0,0 +1,62 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
```
CodrinCotarlanSoftwire, line 2: Terraform documentation suggests to commit this file. See https://stackoverflow.com/questions/67963719/should-terraform-lock-hcl-be-included-in-the-gitignore-file

terraform/postgres.tf
```diff
@@ -0,0 +1,55 @@
...
+  port                        = var.rds_config_port
+  backup_retention_period     = var.rds_config_backup_retention_period
+  backup_window               = var.rds_config_backup_window
+  vpc_security_group_ids      = [aws_security_group.allow_postgres_connection.id]
+  db_subnet_group_name        = module.vpc.database_subnet_group_name
+  storage_encrypted           = var.rds_config_storage_encrypted
+  publicly_accessible         = var.rds_config_publicly_accessible
+  allow_major_version_upgrade = var.rds_config_allow_major_version_upgrade
+  multi_az                    = var.rds_config_multi_az
+  skip_final_snapshot         = var.rds_config_skip_final_snapshot
```
CodrinCotarlanSoftwire, line 47: This may be a topic for the future, but we should discuss sensible default values for each of these, and what the dev / preprod / prod values should be.

terraform/provider.tf
```diff
@@ -1,30 +1,27 @@
...
+    }
+
+    null = {
+      source  = "hashicorp/null"
+      version = "3.1.1"
+    }
+
+  }
+
+  backend "s3" {}
```
CodrinCotarlanSoftwire, line 22: Declared in .tfbackend files

terraform/vpc.tf
```diff
@@ -0,0 +1,47 @@
...
+  // Keeping the 'registry.terraform.io/' prefix so Rider can find the module for autocompletion purposes
+  source  = "registry.terraform.io/terraform-aws-modules/vpc/aws"
+  version = "3.14.2"
+
+  name = join("-", ["gpg-application-vpc", var.env])
+  cidr = "10.0.0.0/16"
+
+  azs              = slice(data.aws_availability_zones.available.zone_ids, 0, 2)
+  public_subnets   = [cidrsubnet("10.0.0.0/16", 4, 0), cidrsubnet("10.0.0.0/16", 4, 1)]
+  database_subnets = [cidrsubnet("10.0.0.0/16", 4, 2), cidrsubnet("10.0.0.0/16", 4, 3)]
```
davidm-m, line 11: This looks worryingly like our database subnets are a subset of our public subnets.

terraform/vpc.tf
```diff
@@ -0,0 +1,47 @@
...
+  database_subnet_assign_ipv6_address_on_creation = false
+
+  public_subnet_ipv6_prefixes   = [0, 1]
+  database_subnet_ipv6_prefixes = [2, 3]
+
+  enable_dns_hostnames = true
+  enable_dns_support   = true
+
+  enable_nat_gateway                     = false
+  single_nat_gateway                     = true
```
davidm-m, line 25: Don't these contradict each other? I don't think we'll be creating any NAT gateways with `enable_nat_gateway` set to `false`

davidm-m (changes requested):
Got a few concerns, but you guys know more than me about Terraform so feel free to argue with me on them.

terraform/variables.tf
```diff
@@ -9,4 +9,104 @@ variable "example_s3_versioning_enabled" {
...
+variable "rds_config_publicly_accessible" {
+  type        = bool
+  default     = true
+  description = "Specifies if the database is publicly accessible"
+}
+
+variable "rds_config_allow_major_version_upgrade" {
+  type        = bool
+  default     = true
+  description = "Specifies if the database can update major versions e.g. 11 -> 12"
```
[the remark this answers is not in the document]
LadunOmideyiSoftwire: I agree it's risky to upgrade so we can turn this to no. I think we should be able to pop a cloudwatch alarm into the terraform config to give us notice of a major version change so that at least we are aware of upgrades.

terraform/vpc.tf
CodrinCotarlanSoftwire: database_subnets is a subset of subnets, and is private by default.

terraform/postgres.tf
```diff
@@ -0,0 +1,55 @@
...
+
+// Outgoing rules
+resource "aws_security_group_rule" "postgres_out" {
+  security_group_id = aws_security_group.allow_postgres_connection.id
+  type              = "egress"
+  from_port         = 0
+  to_port           = 0
+  protocol          = "-1"
+  cidr_blocks       = ["0.0.0.0/0"]
+}
```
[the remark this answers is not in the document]
davidm-m: Why do we want our db to respond anywhere? Surely it should only be talking to connections from our VPC? Or we expecting to connect directly to the database from elsewhere?

davidm-m (changes requested):
Still concerned with allowing universal egress on the db - by default we should be locking down connections (in and out) as much as possible, so if we don't need the db to connect to everywhere we shouldn't let it.

CodrinCotarlanSoftwire:
Since, as you mention, we do not plan on connecting to the database from anything else except the app I believe it is fine to restrict its entire traffic inside he VPC, ingress and egress.

davidm-m (approved):
Lovely stuff.

PR #465, opened 2022-08-23
GPG-461: Add pending message to email address in manage account

CodrinCotarlanSoftwire:


GenderPayGap.WebUI/Controllers/Account/ManageAccountController.cs
```diff
@@ -24,6 +27,15 @@ public IActionResult ManageAccountGet()
...
 
+            // Remove the email change request if it is expired
+            if (currentUser.NewEmailAddressRequestDate.HasValue
+                && currentUser.NewEmailAddressRequestDate.Value.AddDays(Global.EmailVerificationExpiryDays) < VirtualDateTime.Now)
+            {
+                currentUser.NewEmailAddress = null;
+                currentUser.NewEmailAddressRequestDate = null;
+                dataRepository.SaveChanges();
+            }
+
```
CodrinCotarlanSoftwire, line 38: This seems the most reasonable place to check whether the  address expired, however it's a get method. The alternative is to never reset these, but instead perform that check and pass the view model a boolean flag whether to use the current email or the new one for display. Thoughts?

PR #467, opened 2022-08-23
GPG-806: Add character limits to all reason boxes

CodrinCotarlanSoftwire:


davidm-m (approved):
Looks good, thanks!

Do you know how much of this (if any) will need changing under the new design system?

PR #468, opened 2022-08-24
GPG-99: CSV Generation, controllers and front end changes.

Airk0n:
**Changes:**
Additional heading and downloads on the `admin./downloads` page.
image

The additional downloads are duplicates of the the 3 above however 4 new columns are added to the CSV: first name, second name, email address and date/time of the last user to log in who is associated with a given organisation.
image

These new documents have a different file name when downloaded.
image


davidm-m (changes requested):
I don't think you've pushed all your changes here, `EhrcController` is definitely not in a state where it will compile.

davidm-m (changes requested):
One missing return, otherwise perfect.

davidm-m (approved):

PR #469, opened 2022-09-02
Gpg 867 1 elastic beanstalk

LadunOmideyiSoftwire:
This contains:

- The vpc and elastic beanstalk application
- The manifest json file  _NOTE: the Build Gender Pay Gap - AWS pipeline won't succeed until this is merged because it's necessary (along with the archive file) for azure devops to deploy elastic beanstalk_. 
- Environment specific backends
- Additional providers

### DIFF
--- a/terraform/dev-app.tfvars
+++ b/terraform/dev-app.tfvars
@@ -1,4 +1,4 @@
-﻿Instance_type          = "t2.medium"
+﻿Instance_type          = "t2.small"
 elb_instance_profile   = "aws-elasticbeanstalk-ec2-role"
 elb_instance_min_size  = 1
 elb_instance_max_size  = 2
--- a/terraform/dev-app.tfvars
+++ b/terraform/dev-app.tfvars
@@ -1,0 +1,9 @@
+﻿Instance_type          = "t2.medium"
+elb_instance_profile   = "aws-elasticbeanstalk-ec2-role"
+elb_instance_min_size  = 1
+elb_instance_max_size  = 2
+tier                   = "WebServer"
+elb_scheme             = "internet facing"
+elb_load_balancer_type = "application"
+solution_stack_name    = "64bit Amazon Linux 2 v2.3.4 running .NET Core"
+elb_matcher_http_code  = "200"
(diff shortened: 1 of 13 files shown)


terraform/elasticbeanstalk.tf
```diff
@@ -0,0 +1,73 @@
...
+
+// Archive file 
+data "aws_s3_object" "gpg-archive-zip" {
+  bucket = data.aws_s3_bucket.gpg-application-version-storage.id
+  key    = "publish.zip"
+}
+
+// Application
+resource "aws_elastic_beanstalk_application" "gpg-application" {
+  name        = join("-",["gpg-application", var-env])
```
davidm-m, line 14: Is this supposed to be `var-env` or `var.env`?

terraform/dev-app.tfvars
```diff
@@ -0,0 +1,9 @@
+﻿Instance_type          = "t2.medium"
```
davidm-m, line 1: Are we likely to actually need a medium here? Have we looked into whether we could get away with a small on dev and pre-prod?

terraform/main.tf
```diff
@@ -1,15 +1,6 @@
...
-  bucket = "gpg-bucket-${terraform.workspace}"
-  acl    = "private"
-
-  versioning = {
-    enabled = true
-  }
-
+data "aws_availability_zones" "available" {
+  state = "available"
 }
```
davidm-m, line 6: Showing how little I know about terraform here, but do we need to do more than this in the `main.tf` file? Or does it just detect there are more `.tf` files and run them all?

davidm-m (changes requested):
A typo and a couple of questions. Overall looks very good, the separation of types of resources into multiple files is really nice!

terraform/elasticbeanstalk.tf
LadunOmideyiSoftwire: Definitely the latter!

terraform/main.tf
LadunOmideyiSoftwire: It detects all .tf files in the root terraform folder so we should be fine.

terraform/dev-app.tfvars
LadunOmideyiSoftwire: We should be able to get away with a small and I've made the changes.

davidm-m (approved):
Nice!

PR #473, opened 2022-09-21
Gpg 870 ipdeny to waf

CodrinCotarlanSoftwire:
in Infrastructure/gpg-ipdeny are files which details how the ipdeny service was created. I looked through it and implemented the rules in there in the aws resource, notably the whitelist and blacklist.


terraform/waf.tf
```diff
@@ -0,0 +1,122 @@
...
+  addresses          = [for ip in split("\n", file("EHRCDownload-IP-Whitelist.txt")): trimspace(ip)]
+}
+
+resource "aws_wafv2_ip_set" "blacklisted_ips" {
+  name               = "blacklisted-ips"
+  description        = "Blacklisted IPs. These IPs cannot connect to the website."
+  scope              = "REGIONAL"
+  ip_address_version = "IPV4"
+  addresses          = [for ip in split("\n", file("GPG-IP-Denylist.txt")) : trimspace(ip)]
+}
```
CodrinCotarlanSoftwire, line 25: These files will be downloaded by the azure pipeline

terraform/waf.tf
```diff
@@ -0,0 +1,122 @@
...
+          not_statement {
+            statement {
+              ip_set_reference_statement {
+                arn = aws_wafv2_ip_set.ehrc_whitelisted_ips.arn
+              }
+            }
+          }
+        }
+      }
+    }
```
CodrinCotarlanSoftwire, line 108: Block requests which are sent to download?p... and are NOT from the whitelist

terraform/waf.tf
```diff
@@ -0,0 +1,122 @@
+﻿resource "aws_wafv2_regex_pattern_set" "ehrc_protected_request_address" {
+  name        = "ehrc-rotected-request-address"
+  description = "Regex of the endpoint used by ehrc. Of the form .../download?p=filename"
+  scope       = "REGIONAL"
+
+  regular_expression {
+    regex_string = "^/download(/)?[?]p=(.*)$"
```
davidm-m, line 7: The `/` might need to be escaped in this. I would also normally use `\` to escape a special character (i.e. `\?` instead of `[?]`)

terraform/waf.tf
davidm-m: What's the process for updating these files? I know they haven't changed in a while, but I think now's a good time to make sure we're considering that so they can be updated easily in the future.

davidm-m (changes requested):
Couple of small queries, overall looks good!

terraform/waf.tf
CodrinCotarlanSoftwire: The files are on keeper. I'm not aware of any particular way of updating them. I suppose one can download them from keeper, make the changes, and then upload them back on keeper and azure.

davidm-m: Hmm ok. I think because of the time pressure and the fact these files don't change often we can leave it like this.
When it comes to documentation though, I think there should be a big "How to update the IP blacklist/whitelist" heading with detailed instructions on where the files can be found, how to replace them in Azure, and how to make sure those changes are deployed. Just something to remember when we get to that point 🙂

terraform/waf.tf
CodrinCotarlanSoftwire: I've tried to do it, and played around with escaping characters but terraform won't accept escaping neither / nor ?, so I don't know if I can do it in another way.

davidm-m: Weird. As long as this works I'm happy with it then, but I'm surprised you can't escape them with a backslash.

davidm-m (approved):

PR #474, opened 2022-09-22
Gpg 905 mobile list components

Airk0n:
### **Changes:**
 - Table replaced with component library table.
 - Separation between mobile and desktop views.
 - Coloured badges component (reliant on enum).
 - New property in Return for getting the badge required for the report (presumably we can use this across the site).
 - Date that an employer joined the site added.
 - Styling and wording differences between desktop and mobile views.
 
###  Desktop:
image

### Mobile:
image


GenderPayGap.WebUI/GenderPayGap.WebUI.csproj
```diff
@@ -24,6 +24,7 @@
 
   <ItemGroup>
     <Content Remove="Views\Organisation\ManageOrganisationBreadcrumbs.cshtml" />
+    <Content Remove="wwwroot\styles\pages\_employer-reports-table.scss" />
```
davidm-m, line 27: We're not doing the same thing with files in that folder as far as I can tell, do you know why this is here?

davidm-m (changes requested):
Just one question, otherwise looks great!

GenderPayGap.WebUI/GenderPayGap.WebUI.csproj
Airk0n: Not a clue @CodrinCotarlanSoftwire?

CodrinCotarlanSoftwire: I believe we've added a custom style file with a class to hide the tables on desktop / mobile, and I noticed there was a reference to `_employer_reports_table.scss`. However the file doesn't exist, so I removed it from an imports file.

davidm-m: First off, it does exist 😛
image
I presume it's here in the `.csproj` because it's not being used anywhere, so it's automatically being excluded from the build? I think we need to add that import back in or actually delete this file if its style isn't being used.

CodrinCotarlanSoftwire: I was thinking on another file that we indeed removed. I don't know why this one is in a 'Remove' field in csproj...

Airk0n: Removed

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Parts/_ReportStatusBadge.cshtml
```diff
@@ -1,10 +1,80 @@
...
+    Func<object, object> html = null;
+    string htmlClass = "";
+
+
+    switch(Model.ReportStatus)
+    {
+        case ReportStatusBadgeType.Due:
+            if (Model.Desktop)
+            {
+                html = @<text>DUE @Model.DateText.ToUpper()</text>;
```
davidm-m, line 16: `<text>` isn't a real HTML tag as far as I know. Is there a particular reason you're using it here?

davidm-m (changes requested):

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Parts/_ReportStatusBadge.cshtml
Airk0n: It's essentially just being used for string interpolation here, we couldn't give these badges text as the formatting was incorrect and the classes are baked into the library component.

davidm-m (approved):

PR #475, opened 2022-09-26
Gpg 869 cache

LadunOmideyiSoftwire:
Stripped down to most basic config - 1.34gb (slight downgrade from 1.5gb), single node, no failover.

Also making slight tweaks to variables.

### DIFF
--- a/terraform/prod-app.tfvars
+++ b/terraform/prod-app.tfvars
@@ -11,8 +11,6 @@ env = "prod"
 
 #region ElasticBeanstalk configuration
 
-Instance_type          = "t2.medium"
-instance_type          = "t2.medium"
 instance_type          = "t2.medium"
 elb_instance_profile   = "aws-elasticbeanstalk-ec2-role"
 elb_instance_min_size  = 1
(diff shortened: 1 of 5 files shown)


terraform/prod-app.tfvars
```diff
@@ -20,5 +20,6 @@ elb_scheme             = "internet facing"
 elb_load_balancer_type = "application"
 solution_stack_name    = "64bit Amazon Linux 2 v2.3.4 running .NET Core"
 elb_matcher_http_code  = "200"
+cache_port = 7410
```
davidm-m, line 23: Is there a particular reason we're using different ports in each environment?

davidm-m (approved):
Looks fine to me! Just one question about the ports.

terraform/prod-app.tfvars
davidm-m: I'm not saying this needs to be changed, just wondering what the reasoning behind it is. Generally the more similar your different environments are, the less likely you are to run into weird issues in the future.

PR #476, opened 2022-09-27
Gpg 909/910 add js filter

Airk0n:
**Overview:**
The purpose of this ticket is to add a filter that allows users to see a only a specific years details and then return to seeing all years with a "show all option".
This is to be done using a `<Select>`  element and is to only be visible to users with JavaScript enabled on their browsers.
This is to be visible on both mobile and desktop(covered in 2 tickets).

**Changes:**
- reorganisation, text removal and restructuring of area above the years (as per the Figma designs - https://www.figma.com/file/vPKrhKrWG6z762gFgYgkvR/GPG-Dev-Handover?node-id=108%3A120).
- Select, hint and label elements created and appended to div via javascript.
- Hidden <input> passes year options as json where they are picked up by the js and used to create options.
- Options are then compared with textContent in `<h3>` tags.
- the table rows are then selected by looking at the parent nodes until a `<tr>` tag is reached.

**Select dropdown:**
image

**Specific year selected:**
image

### DIFF
--- a/GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
+++ b/GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
@@ -349,10 +349,7 @@
         selectList.id = "filterYearSelect";
         selectList.addEventListener("change",function (e){
             for (var i = 0; i < elements.length; i++){
-                var tableRow = elements[i];
-                while (tableRow.tagName.toLowerCase() !== "tr"){
-                    tableRow = tableRow.parentElement;    
-                }
+                var tableRow = elements[i].closest("tr");
                 tableRow.hidden = true;
                 
                 if (elements[i].innerHTML.includes(e.target.value) || e.target.value === "Show All"){
@@ -376,13 +373,12 @@
         selectList.appendChild(showAll);
         
         var arrayOfYears = JSON.parse(document.getElementById("selectOptionsFromReports").value);
-        for (var i = 1; i < arrayOfYears.length; i++){
+        for (var i = 0; i < arrayOfYears.length; i++){
              var option = document.createElement("option");
              option.value = arrayOfYears[i];
              option.text = arrayOfYears[i];
              selectList.appendChild(option);
         }
         
-        selectList.selectedIndex = 0;
     </script>
 }
--- a/GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
+++ b/GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
@@ -2,6 +2,7 @@
 @using GenderPayGap.WebUI.Models.Shared
 @using GenderPayGap.Core
 @using GenderPayGap.Core.Helpers
+@using GenderPayGap.WebUI.Models.Search
 @using GovUkDesignSystem
 @using GovUkDesignSystem.GovUkDesignSystemComponents
 @using Newtonsoft.Json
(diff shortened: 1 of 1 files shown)


GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
```diff
@@ -307,6 +334,55 @@
...
+        
+        var selectList = document.createElement("select");
+        selectList.className = "govuk-select govuk-!-margin-bottom-3";
+        selectList.id = "filterYearSelect";
+        selectList.addEventListener("change",function (e){
+            for (var i = 0; i < elements.length; i++){
+                var tableRow = elements[i];
+                while (tableRow.tagName.toLowerCase() !== "tr"){
+                    tableRow = tableRow.parentElement;    
+                }
```
davidm-m, line 355: Do we need to support IE? If not, you can simplify this to `var tableRow = elements[i].closest("tr")` ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest))

davidm-m: @LadunOmideyiSoftwire the first bit is probably a question for you.

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
```diff
@@ -307,6 +334,55 @@
...
+        parentForSelect.appendChild(selectHint);
 
+        var showAll = document.createElement("option");
+        showAll.value = "Show All";
+        showAll.text = "Show All";
+        showAll.selected = true;
+        selectList.appendChild(showAll);
+        
+        var arrayOfYears = JSON.parse(document.getElementById("selectOptionsFromReports").value);
+        for (var i = 1; i < arrayOfYears.length; i++){
```
davidm-m, line 379: Is `i = 1` deliberate?

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
```diff
@@ -307,6 +334,55 @@
...
+             var option = document.createElement("option");
+             option.value = arrayOfYears[i];
+             option.text = arrayOfYears[i];
+             selectList.appendChild(option);
+        }
+        
+        selectList.selectedIndex = 0;
```
davidm-m, line 386: Is this necessary? I would assume that's the default.

davidm-m (changes requested):
Nice stuff. I've got a few small suggestions.

It would be really nice to see a screenshot of this to make sure it matches designs 🙂

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
Airk0n: It was intentional but incorrect, I was offsetting for the "show all" but that's not required.

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
Airk0n: correct. It's unnecessary.

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
Airk0n: Firstly, this wasn't me intentionally supporting IE, I was just unaware of that option. But I would presume IE should be supported.

davidm-m: Yeah I didn't think this was a deliberate choice for IE compatibility, just wanted to make sure we didn't need to support it before I suggested adding a feature that isn't in IE.

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
```diff
@@ -62,14 +64,48 @@
...
+                                <div id="filteringArea">
+                                    <input type="hidden" id="selectOptionsFromReports"
+                                           value=@JsonConvert.SerializeObject(
+                                                     organisation.GetRecentReports(
+                                                         Global.ShowReportYearCount)
+                                                         .Select(
+                                                             report => report.GetReportingPeriod())
+                                                         .ToArray())>
+                                </div>
+                            }
```
davidm-m, line 105: Should this code block have ended at line 90? It feels weird we're writing html straight into a code block.
I also think the `GovUkSelect` might need an `await`

PR #484, opened 2022-10-03
Update vcap services in terraform

CodrinCotarlanSoftwire:


terraform/elasticbeanstalk.tf
```diff
@@ -92,7 +96,7 @@ resource "aws_elastic_beanstalk_environment" "gpg-elb-environment" {
   setting {
     namespace = "aws:elasticbeanstalk:application:environment"
     name      = "VCAP_SERVICES"
-    value     = var.ELB_VCAP_SERVICES
+    value     = "{\"postgres\":[{\"credentials\":{\"password\":\"${aws_db_instance.gpg-dev-db.password}\",\"port\":${aws_db_instance.gpg-dev-db.port},\"name\":\"${aws_db_instance.gpg-dev-db.name}\",\"host\":\"${aws_db_instance.gpg-dev-db.endpoint}\",\"username\":\"${aws_db_instance.gpg-dev-db.username}\"},\"name\":\"${aws_db_instance.gpg-dev-db.name}\"}],\"redis\":[{\"credentials\":{\"password\":\"\",\"tls_enabled\":true,\"port\":${aws_elasticache_cluster.redis-cluster.port},\"host\"\"${aws_elasticache_cluster.redis-cluster.cache_nodes[0].address}\"},\"name\":\"gpg-${var.env}-cache\"}],\"aws-s3-bucket\":[{\"name\":\"${aws_s3_bucket.gpg-filestorage.bucket}\",\"credidentials\":{\"aws_region\":\"${var.aws_region}\",\"bucket_name\":\"${aws_s3_bucket.gpg-filestorage.bucket}\", \"aws_access_key_id\":\"${var.AWS_ACCESS_KEY_ID}\",\"aws_secret_access_key\":\"${var.AWS_SECRET_ACCESS_KEY}\"}}]}"
```
CodrinCotarlanSoftwire, line 99: The code currently uses this one environment value to parse most of it's connections, so for now in order to reduce the code changes to 0 we'll do it this way.

terraform/variables.tf
```diff
@@ -212,4 +208,18 @@ variable "ELB_REPORTING_START_YEARS_TO_EXCLUDE_FROM_LATE_FLAG_ENFORCEMENT" {
...
+variable "AWS_ACCESS_KEY_ID" {
+  type        = string
+  description = "AWS access key id. Set as an environment variable."
+}
+
+variable "AWS_SECRET_ACCESS_KEY" {
+  type        = string
+  description = "AWS secret access key. Set as an environment variable."
+}
+
```
CodrinCotarlanSoftwire, line 224: Needed for the s3 bucket in VCAP_SERVICES

terraform/elasticbeanstalk.tf
davidm-m: [Terraform has a jsonencode function](https://www.terraform.io/language/expressions/strings#generating-json-or-yaml) that should let you set this out more neatly

davidm-m (changes requested):
If you could switch that string for a jsonencode block that would make this much easier to read and maintain, otherwise looks good!

terraform/elasticbeanstalk.tf
CodrinCotarlanSoftwire: Yes, I agree, updated the code to use that! Looks much more maintainable

davidm-m (approved):
Nice 😄

PR #494, opened 2022-10-11
GPG-909: unneeded select object removed, formatting cleanup.

Airk0n:
Removal of unnecessary component: There was 2 selects, one in C# one in JS. the C# one has been removed and the spacing has been fixed in the code.


davidm-m (approved):
Thanks!

PR #495, opened 2022-10-18
Gpg 911 view reports pagination

Airk0n:
The purpose of this ticket was to set a limit of 10 table rows per page, after which point pagination appears allowing the user to navigate pages whilst not filling the screen.

This applies to both mobile and desktop.

image

image

The current items per row is 10 meaning that at this point the pagination will not show on the live site, these images are tested with a row limit of 3.


GenderPayGap.WebUI/Content/Pagination.css
```diff
@@ -155,3 +155,156 @@
...
+    -webkit-box-direction: normal;
+    flex-direction: column;
+    -ms-flex-align: center;
+    -webkit-box-align: center;
+    -ms-flex-wrap: wrap;
+    flex-wrap: wrap;
+}
+
+.govuk-pagination--block .govuk-pagination__link:not(:focus) {
+    text-decoration: none;
```
davidm-m, line 309: Is this all from gov uk? I'd assume it's already included in the design library

GenderPayGap.WebUI/Controllers/ViewingController.cs
```diff
@@ -323,14 +323,31 @@ public IActionResult Employer(string employerIdentifier)
             ReportBackUrl = null;
 
             ViewBag.BasketViewModel = new CompareBasketViewModel {CanAddEmployers = true, CanViewCompare = true};
+            
+            int totalEntries = organisationLoadingOutcome.Result.GetRecentReports(Global.ShowReportYearCount).Count() + 1; // Years we report for + the year they joined
+            int maxEntriesPerPage = 10;
+            int totalPages = (int)MathF.Ceiling(totalEntries / maxEntriesPerPage);
```
davidm-m, line 329: Is there a particular reason we're using `int` instead of `var`?

davidm-m, line 329: I think you've got some Unity showing through 😄
I'd never heard of `MathF` before, looks like it's a Unity library that's a copy of `Math` except it uses floats instead of doubles. Normally we'd just use `Math`.

GenderPayGap.WebUI/Views/Components/PaginationLinks/GpgPaginationLinksViewModel.cs
```diff
@@ -1,17 +1,17 @@
 namespace GenderPayGap.WebUI.Views.Components.PaginationLinks
 {
-    public class PaginationLinksViewModel
```
davidm-m, line 3: The design library has a pagination component. Can we get rid of this and use that instead?

GenderPayGap.WebUI/Views/Viewing/EmployerDetails/Employer.cshtml
```diff
@@ -62,235 +62,274 @@
...
+                    Id = "WhoNeedsToReport",
+                    LinkText = "Who is required to report, and the deadlines",
+                    SummaryPartial = "EmployerDetails/WhoNeedsToReport"
+                })
+                <p>
+                    For example, if the snapshot date is 5 April 2017 but the deadline is 4 April 2018 then the reporting period is 2017 - 2018.
+                </p>
 
-                    </div>
-                </div>
```
davidm-m, line 93: It looks like a whole bunch of stuff in this file got unindented by two tabs for no reason I can see and that's making it really hard to tell what's actually changed from the diff. Could you fix the indentation please?

davidm-m (changes requested):

davidm-m (dismissed):

davidm-m (dismissed):

davidm-m (approved):

PR #498, opened 2022-10-26
Gpg 759 manage reports page

Airk0n:
This page is for users who are logged in and are associated with a company. This gives them a sort of dashboard where they can move onto other actions such as completing a report, editing an old report or viewing all reports.

### Changes:
-  There was one large table which held info for every reporting year on one page, we now have 2 tables, 1) latest due report, 2) missing reports.
image
 - Company registered address moved to bottom of the page and styled as per the rest of the page.
image
 - Link to view all reports page created (though not linked up yet).
image

Note: I moved the "registered users" table over to using a design library component as I had to make a visual change to it and saw an opportunity to improve the maintainability of the code.

### DIFF
--- a/GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
+++ b/GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
@@ -60,32 +60,12 @@
         
         @{
             await Html.RenderPartialAsync("OrganisationReportsTable", Model);
-        }
-        
-        <h2 class="govuk-heading-l">Registered users</h2>
 
-        @{
-            var usersRegisteredToReportForOrganisation = Model.Organisation.UserOrganisations.Where(uo => uo.PINConfirmedDate.HasValue).Select(uo => uo.User).Distinct().ToList();
-        }
-        @if (usersRegisteredToReportForOrganisation.Count == 0)
-        {
-            <p class="govuk-body">
-                You are the only person registered to report for this employer.
-            </p>
-            <p class="govuk-body">
-                If you remove yourself:
-            </p>
-            <ul class="govuk-list govuk-list--bullet">
-                <li>You will not be able to report for this employer</li>
-                <li>Someone else must register this employer to report - this can take up to a week</li>
-                <li>Your account will remain open</li>
-            </ul>
-        } else
-        {
-            <p class="govuk-body">
-                The following people are registered to report gender pay gap information for this employer. 
-            </p>
-
+            <div class="govuk-!-margin-bottom-6">
+                <p>
+                    <a href="" class="govuk-!-font-size-19 govuk-link govuk-!-font-weight-bold">See all reports</a> <!--TODO: Point this link to the manage all reports page when that is created-->
+                </p>
+            </div>
             await Html.RenderPartialAsync("UsersRegisteredToReportForOrganisation", Model);
         }
         
--- a/GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
+++ b/GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
@@ -41,19 +41,7 @@
                 for @(Model.Organisation.OrganisationName)
             </span>
         </h1>
-        
-        <dl>
-            <dt class="govuk-body govuk-!-font-weight-bold">
-                Registered Address
-            </dt>
-            <dd class="govuk-caption-m govuk-!-font-size-16 govuk-!-margin-left-0">
-                @foreach (string addressLine in Model.Organisation.GetLatestAddress().GetAddressLines())
-                {
-                    @(addressLine)<br/>
-                }
-            </dd>
-        </dl>
-        
+
         <details class="govuk-details" data-module="govuk-details">
             <summary class="govuk-details__summary">
                 <span class="govuk-details__summary-text">
@@ -70,7 +58,6 @@
             </div>
         </details>
         
-
         @{
             await Html.RenderPartialAsync("OrganisationReportsTable", Model);
         }
@@ -101,6 +88,17 @@
 
             await Html.RenderPartialAsync("UsersRegisteredToReportForOrganisation", Model);
         }
+        
+        <h2 class="govuk-heading-l">
+            Registered Address
+        </h2>
+        <p class="govuk-body">
+            @foreach (string addressLine in Model.Organisation.GetLatestAddress().GetAddressLines())
+            {
+                @(addressLine)<br/>
+            }
+        </p>
+
 
     </div>
 </div>
--- a/GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
+++ b/GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
@@ -63,7 +63,7 @@
 
             <div class="govuk-!-margin-bottom-6">
                 <p>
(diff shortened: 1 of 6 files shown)


GenderPayGap.WebUI/Views/ManageOrganisations/ManageOrganisation.cshtml
```diff
@@ -70,37 +58,27 @@
             </div>
         </details>
         
-        <h2 class="govuk-heading-l">Manage reports</h2>
         @{
             await Html.RenderPartialAsync("OrganisationReportsTable", Model);
-        }
```
Airk0n, line 76: Moved into the partial

GenderPayGap.WebUI/Views/ManageOrganisations/UsersRegisteredToReportForOrganisation.cshtml
```diff
@@ -1,35 +1,130 @@
...
+                new TableCellViewModel
+                {
+                    Html = @<text>
+                               @userRegisteredToReport.Fullname
+                               @if (userRegisteredToReport.UserId == Model.User.UserId)
+                               {
+                                   <span>(You)</span>
+                               }
+                               <br/>
+                            </text>
```
davidm-m, line 73: Can this be wrapped in a `<p>` instead of using `<br/>`? Is that even needed? The next cell is just a naked `<a>` which makes me think it might not be.

GenderPayGap.WebUI/Views/ManageOrganisations/UsersRegisteredToReportForOrganisation.cshtml
```diff
@@ -1,35 +1,130 @@
...
+    }
+    
+}
+
+<h2 class="govuk-heading-l">Registered users</h2>
+    @{
+        var usersRegisteredToReportForOrganisation = Model.Organisation.UserOrganisations.Where(uo => uo.PINConfirmedDate.HasValue).Select(uo => uo.User).Distinct().ToList();
+    }
+
+    @if (usersRegisteredToReportForOrganisation.Count == 0)
```
davidm-m, line 95: Is this the correct logic? Surely it's always at least one, the person viewing this page?

GenderPayGap.WebUI/Views/ManageOrganisations/UsersRegisteredToReportForOrganisation.cshtml
```diff
@@ -1,35 +1,130 @@
...
+            {
+                new TableCellViewModel
+                {
+                    Html = @<text>
+                               <p class="govuk-!-margin-top-1 govuk-!-margin-bottom-1">
+                                   @userRegisteredToReport.Fullname
+                                   @if (userRegisteredToReport.UserId == Model.User.UserId)
+                                   {
+                                       <span>(You)</span>
+                                   }
```
davidm-m, line 38: Feels like there must be some de-duplication we can do with bits of html like this. It's definitely not worth making a whole partial but a private function so that the desktop and mobile code share as much as possible might be worth it. What do you think?

davidm-m (changes requested):
Let me know if you want to discuss any of these comments, I don't know if I explained all of them well.
Also, could you add some screenshots of this on mobile?

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
```diff
@@ -1,72 +1,267 @@
...
+                AddRowMobile(mobileMissingReportingPeriodRows, organisationReturn);
+                break;
+            case ReportTag.NotRequired:
+                if (organisationReturn.ReportingYear == currentReportingYear)
+                {
+                    AddRowDesktop(desktopCurrentReportingPeriodRows, organisationReturn);
+                    AddRowMobile(mobileCurrentReportingPeriodRows, organisationReturn);
+                }
+                break;
+        }
```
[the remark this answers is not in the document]
Airk0n: Yeah I agree, this was clearly me going through condition by condition and not thinking about what the end result of that actually was.
I'll go with the else if as there's only 2 cases.

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
```diff
@@ -1,72 +1,267 @@
...
+        };
+        rowList.Add(newRow);
+    }
+}
+
+<h2 class="govuk-heading-l gpg-govuk-hideOnMobile">Manage reports</h2>
+<h2 class="govuk-heading-l gpg-govuk-hideOnDesktop govuk-!-margin-bottom-0">Manage reports</h2>
+@{
+    var missingReports = desktopMissingReportingPeriodRows.Count > 0;
+    var mobileReportingPeriodsClasses = missingReports ? "gpg-govuk-hideOnDesktop govuk-!-margin-bottom-9" : "gpg-govuk-hideOnDesktop govuk-!-margin-bottom-1";
```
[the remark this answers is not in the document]
Airk0n: Because there is a link to all reports after the tables, we need to make sure that the final table has a small margin to keep the all reports close to the end of the table. 
Put another way, if there's missing reports this table won't be the last one so we need a big margin.

GenderPayGap.WebUI/Views/ManageOrganisations/UsersRegisteredToReportForOrganisation.cshtml
```diff
@@ -1,35 +1,130 @@
...
+                                   <a class="govuk-link" href="@(Url.Action("RemoveOrganisation", "Organisation", new {orgId = encryptedOrganisationId, userId = encryptedUserId}))">
+                                       Remove user <span class="govuk-visually-hidden">@userRegisteredToReport.Fullname</span>
+                                   </a>
+                               </p>
+                            </text>
+                },
+                new TableCellViewModel
+                {
+                    Text = " "
+                }
```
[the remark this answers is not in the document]
Airk0n: The amount of columns is determined by how many TableCellViewModels are in the list, adding a blank one gives me this column at the end which could have the title "Actions" but it's omitted from the designs.
image

GenderPayGap.WebUI/Views/ManageOrganisations/UsersRegisteredToReportForOrganisation.cshtml
Airk0n: I'm pretty sure it doesn't function as intended however I also do not know how it was intended to work, this code was legacy and moved from the view into this partial.
I will create a ticket for someone to take a look at it after confirming how it's use is intended.

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
```diff
@@ -1,72 +1,267 @@
...
+                if (organisationReturn.ReportingYear == currentReportingYear)
+                {
+                    AddRowDesktop(desktopCurrentReportingPeriodRows, organisationReturn);
+                    AddRowMobile(mobileCurrentReportingPeriodRows, organisationReturn);
+                }
+                break;
+        }
+    }
+
+    void AddRowDesktop(List<TableRowViewModel> rowList, ManageOrganisationDetailsForYearViewModel viewModel)
```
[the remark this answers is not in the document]
Airk0n: Yes, there's a potential for the List.Add to go unnoticed in the function, I'll change that.

davidm-m (commented):
Nice, thanks for making those changes!

PR #500, opened 2022-11-07
Gpg 916 all reports page

Airk0n:
This ticket is concerned with the creation of a new "all reports" page, the idea being the that previous page covered in GPG-759 shows a "dashboard" of actionable reports whereas the the "all reports" page shows all previous reports including the unactionable ones and serves as an archive, this includes pagination and yearly filtering.

Image showing the design for desktop- 
image

Image showing my implementation on desktop
image
Note: a change was agreed with the designer that we could reuse badges seen on the viewing pages, this is why the report status badges look different or are worded differently.

Image showing mobile design - 
image

Image showing mobile implementation - 
image

Note: a design change was agreed where the "draft report" button was moved below the badge so as to match the government design standards of having only single column tables on mobile devices.

Pagination is not visible as there needs to be more than 10 reports available for the next and previous buttons to appear.

### DIFF
--- a/GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
+++ b/GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
@@ -1,10 +1,16 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using GenderPayGap.Core;
 using GenderPayGap.Core.Classes;
+using GenderPayGap.Core.Classes.ErrorMessages;
+using GenderPayGap.Core.Classes.Logger;
 using GenderPayGap.Core.Interfaces;
+using GenderPayGap.Core.Models;
+using GenderPayGap.Core.Models.HttpResultModels;
 using GenderPayGap.Database;
 using GenderPayGap.Database.Models;
+using GenderPayGap.WebUI.BusinessLogic.Services;
 using GenderPayGap.WebUI.Helpers;
 using GenderPayGap.WebUI.Models.ManageOrganisations;
 using Microsoft.AspNetCore.Authorization;
@@ -68,6 +74,50 @@ namespace GenderPayGap.WebUI.Controllers.ManageOrganisations
             var viewModel = new ManageOrganisationViewModel(organisation, user, allDraftReturns);
 
             return View("ManageOrganisation", viewModel);
+        }
+
+        [HttpGet("{encryptedOrganisationId}/AllReports")]
+        [Authorize(Roles = LoginRoles.GpgEmployer)]
+        public IActionResult AllOrganisationReportsGet(string encryptedOrganisationId, int? page = 1)
+        {
+            long organisationId = ControllerHelper.DecryptOrganisationIdOrThrow404(encryptedOrganisationId);
+            User user = ControllerHelper.GetGpgUserFromAspNetUser(User, dataRepository);
+            ControllerHelper.ThrowIfUserAccountRetiredOrEmailNotVerified(user);
+            ControllerHelper.ThrowIfUserDoesNotHavePermissionsForGivenOrganisation(User, dataRepository, organisationId);
+            
+            var organisation = dataRepository.Get<Organisation>(organisationId);
+            if (OrganisationIsNewThisYearAndHasNotProvidedScopeForLastYear(organisation))
+            {
+                return RedirectToAction("DeclareScope", "Organisation", new { id = encryptedOrganisationId });
+            }
+            
+            // build the view model
+            List<DraftReturn> allDraftReturns =
+                dataRepository.GetAll<DraftReturn>()
+                    .Where(d => d.OrganisationId == organisationId)
+                    .ToList();
+            
+            if (string.IsNullOrWhiteSpace(encryptedOrganisationId))
+            {
+                return new HttpBadRequestResult("Missing employer identifier");
+            }
+            
+            var totalEntries = organisation.GetRecentReports(Global.ShowReportYearCount).Count() + 1; // Years we report for + the year they joined
+            var maxEntriesPerPage = 10;
+            var totalPages = (int)Math.Ceiling((double)totalEntries / maxEntriesPerPage);
+
+            if (page < 1)
+            {
+                page = 1;
+            }
+
+            if (page > totalPages)
+            {
+                page = totalPages;
+            }
+            
+            var viewModel = new AllOrganisationReportsViewModel(organisation, user, allDraftReturns, page, totalEntries, maxEntriesPerPage);
+            return View("AllOrganisationReports", viewModel);
         }
 
         private static bool OrganisationIsNewThisYearAndHasNotProvidedScopeForLastYear(Organisation organisation)
--- a/GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
+++ b/GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
@@ -96,12 +96,7 @@ namespace GenderPayGap.WebUI.Controllers.ManageOrganisations
                 dataRepository.GetAll<DraftReturn>()
                     .Where(d => d.OrganisationId == organisationId)
                     .ToList();
-            
-            if (string.IsNullOrWhiteSpace(encryptedOrganisationId))
-            {
-                return new HttpBadRequestResult("Missing employer identifier");
-            }
-            
+
             var totalEntries = organisation.GetRecentReports(Global.ShowReportYearCount).Count() + 1; // Years we report for + the year they joined
             var maxEntriesPerPage = 10;
             var totalPages = (int)Math.Ceiling((double)totalEntries / maxEntriesPerPage);
--- a/GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
+++ b/GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
@@ -116,7 +116,7 @@ namespace GenderPayGap.WebUI.Controllers.ManageOrganisations
                 page = totalPages;
             }
             
-            var viewModel = new AllOrganisationReportsViewModel(organisation, user, allDraftReturns, page, totalEntries, maxEntriesPerPage);
+            var viewModel = new AllOrganisationReportsViewModel(organisation, user, allDraftReturns, page, totalPages, maxEntriesPerPage);
             return View("AllOrganisationReports", viewModel);
         }
 
--- a/GenderPayGap.WebUI/Models/ManageOrganisations/AllOrganisationReportsViewModel.cs
+++ b/GenderPayGap.WebUI/Models/ManageOrganisations/AllOrganisationReportsViewModel.cs
@@ -1,0 +1,75 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using GenderPayGap.Core.Helpers;
+using GenderPayGap.Database;
+using GenderPayGap.Database.Models;
+using GenderPayGap.WebUI.Models.Admin;
+
+namespace GenderPayGap.WebUI.Models.ManageOrganisations
+{
+    public class AllOrganisationReportsViewModel
+    {
+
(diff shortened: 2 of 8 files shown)


LadunOmideyiSoftwire (changes requested):
Can we change the colour of the "Draft Report" and other text to the light blue colour in the design?

LadunOmideyiSoftwire (changes requested):
Can we increase the spacing below the "Draft report" on the mobile view to reflect the original spacing below the badges? I think it looks a little squashed.

Airk0n:
> 

Links colour changed via "govuk-link" class.
image

Airk0n:
> Can we increase the spacing below the "Draft report" on the mobile view to reflect the original spacing below the badges? I think it looks a little squashed.

Spacing added below the draft report buttons to match the margin below the reports badges (30px)
image

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
```diff
@@ -0,0 +1,329 @@
...
+@using GenderPayGap.Core.Helpers
+@using GenderPayGap.WebUI.Models.ManageOrganisations
+@using GenderPayGap.WebUI.Models.Shared
+@using GovUkDesignSystem
+@using GovUkDesignSystem.GovUkDesignSystemComponents
+@model GenderPayGap.WebUI.Models.ManageOrganisations.AllOrganisationReportsViewModel
+
+@{
+    ViewBag.Title = "View all your employers reports - Gender pay gap service";
+    Layout = "~/Views/GovUkFrontend/GovUkFrontendLayout.cshtml";
```
davidm-m, line 13: Is this layout actually doing anything? I don't think we normally need one.

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
```diff
@@ -0,0 +1,329 @@
...
+    var organisation = Model.Organisation;
+    string encryptedOrganisationId = Encryption.EncryptQuerystring(Model.Organisation.OrganisationId.ToString());
```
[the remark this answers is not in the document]
davidm-m: We should either use it wherever we're currently doing `Model.Organisation` or get rid of it, don't mind which.

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
```diff
@@ -0,0 +1,329 @@
...
+                                      href="@Url.Action("ReportOverview", "ReportOverview", new {encryptedOrganisationId, reportingYear = report.ReportingYear, canTriggerLateSubmissionWarning = true})">
+                                       @report.GetReportLinkText()
+                                   </a>
+                                </text>
+                    },
+                }
+            };
+            return newRow;
+        }
+        TableRowViewModel GenerateMobileRow(AllOrganisationReportsForYearViewModel report)
```
davidm-m, line 147: I'd quite like there to be an extra line here.

GenderPayGap.WebUI/Models/Shared/ReportStatusBadgeViewModel.cs
```diff
@@ -1,13 +1,15 @@
...
+namespace GenderPayGap.WebUI.Models.Shared
 {
 
     public class ReportStatusBadgeViewModel
     {
         public ReportStatusBadgeType ReportStatus { get; set; }
         public string DateText { get; set; }
         public bool Desktop { get; set; }
+        public string DesktopClasses { get; set; }
+        public string MobileClasses  { get; set; }
```
davidm-m, line 12: If we're already telling it whether it's on desktop or not, why do we need to differentiate between desktop and mobile classes? Are we ever going to need to provide both?

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
```diff
@@ -0,0 +1,329 @@
...
+            allDesktopRows.Insert(indexOfCompanyJoined, dateJoinedRow);
+        }
+        else
+        {
+            allMobileRows.Add(dateJoinedRow);
+            allDesktopRows.Add(dateJoinedRow);
+        }
+        
+        var currentPage = Model.CurrentPage ?? 1;
+        var entriesPerPage = Model.EntriesPerPage ?? 10;
```
davidm-m, line 225: Pretty sure these are both set correctly by the controller, do we need the null checking here?

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
```diff
@@ -0,0 +1,329 @@
...
+        var endIndex = startIndex + entriesPerPage > allDesktopRows.Count ? allDesktopRows.Count : startIndex + entriesPerPage;
+                
+        var desktopDisplayRows = new List<TableRowViewModel>();
+        var mobileDisplayRows = new List<TableRowViewModel>();
+                    
+        for (int i = startIndex; i < endIndex; i++)
+        {
+            mobileDisplayRows.Add(allMobileRows[i]);
+            desktopDisplayRows.Add(allDesktopRows[i]);
+        }
```
davidm-m, line 236: LINQ can do this too!
```C#
var desktopDisplayRows = allDesktopRows.Where((row, i) => i >= startIndex && i < endIndex).ToList();
var mobileDisplayRows = allMobileRows.Where((row, i) => i >= startIndex && i < endIndex).ToList();
```

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
```diff
@@ -0,0 +1,329 @@
...
+    selectList.appendChild(showAll);
+    
+    var arrayOfYears = JSON.parse(document.getElementById("selectOptionsFromReports").value);
+    for (var i = 0; i < arrayOfYears.length; i++){
+         var option = document.createElement("option");
+         option.value = arrayOfYears[i];
+         option.text = arrayOfYears[i];
+         selectList.appendChild(option);
+    }
+</script>
```
davidm-m, line 326: Is this exactly the same as on the other page with filtering? If so (or if it's so close we can make it the same) can we split it out into a partial?

davidm-m (changes requested):
Sorry, this ended up being quite a big one! Let me know if you want to chat about any of these comments.

GenderPayGap.WebUI/Controllers/ManageOrganisations/ManageOrganisationsController.cs
```diff
@@ -70,6 +76,50 @@ public IActionResult ManageOrganisationGet(string encryptedOrganisationId)
...
+            
+            var organisation = dataRepository.Get<Organisation>(organisationId);
+            if (OrganisationIsNewThisYearAndHasNotProvidedScopeForLastYear(organisation))
+            {
+                return RedirectToAction("DeclareScope", "Organisation", new { id = encryptedOrganisationId });
+            }
+            
+            // build the view model
+            List<DraftReturn> allDraftReturns =
+                dataRepository.GetAll<DraftReturn>()
```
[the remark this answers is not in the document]
Airk0n: I think it's different, I see some functionality around checking if the DraftReturn is the same as a submitted Return and deleting the draft accordingly.

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
Airk0n: Agreed and added.

GenderPayGap.WebUI/Models/Shared/ReportStatusBadgeViewModel.cs
Airk0n: Yeah, good spot, I've adjusted the this.

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
Airk0n: This was purely because I was in the habit of setting for loops with an `int i` but that can be `var` however I believer I'll be removing this for loop as per another comment anyway - 
image

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
Airk0n: Ahh ok, I did not, That's much better.

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
Airk0n: They are subtly different, in terms of label and the existence of a hint div on the employer page, 
it's debatable if worth doing for these two, I'm thinking if it comes up again in a future ticket then it's worth giving it a view model and a partial?
I am of course happy to do it if you feel it is worth doing at this point.

davidm-m: Maybe a tech debt ticket instead then. Happy to leave for now as long as there's a reminder to improve it somewhere.

GenderPayGap.WebUI/Models/ManageOrganisations/AllOrganisationReportsViewModel.cs
```diff
@@ -0,0 +1,233 @@
...
+                            reportingYear,
+                            allDraftReturns.Where(d => d.SnapshotYear == reportingYear)
+                                .OrderByDescending(d => d.Modified)
+                                .FirstOrDefault()
+                        )
+                    );
+                }
+            }
+
+            return detailsForYears;
```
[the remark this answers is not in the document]
Airk0n: Ahh I see, 
In this case I would say it's harder to read for not much space saved

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
Airk0n: You have to reference a layout to use the `@section BeforeMain` for breadcrumbs so I think I have to specify it.

GenderPayGap.WebUI/Views/ManageOrganisations/AllOrganisationReports.cshtml
```diff
@@ -0,0 +1,329 @@
...
+
+        var yearsForSelectComponent = new List<string>();
+        var allDesktopRows = new List<TableRowViewModel>();
+        var allMobileRows = new List<TableRowViewModel>();
+        var allReports = Model.GetOrganisationDetailsForYears();
+        foreach (var report in allReports)
+        {
+            allDesktopRows.Add(GenerateDesktopRow(report));
+            allMobileRows.Add(GenerateMobileRow(report));
+        }
```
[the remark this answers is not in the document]
Airk0n: agreed

GenderPayGap.WebUI/Models/ManageOrganisations/AllOrganisationReportsViewModel.cs
```diff
@@ -0,0 +1,233 @@
...
+        }
+    }
+    
+    public class AllOrganisationReportsForYearViewModel
+    {
+
+        public int ReportingYear { get; }
+
+        private readonly Database.Organisation organisation;
+        private readonly DraftReturn draftReturnForYear;
```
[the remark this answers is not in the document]
Airk0n: It's null check gets used here - 
image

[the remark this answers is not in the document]
davidm-m: Yes, but that doesn't mean we need to store `draftReturnForYear` does it? It's enough to store `hasDraftReturnForYear` in the constructor.

LadunOmideyiSoftwire (approved):

davidm-m (approved):

PR #501, opened 2022-11-15
Gpg 918 start report page

Airk0n:
This ticket is concerned with the creation of the page that starts the new reporting journey.

Design screenshot:
image

Implementation:
image


GenderPayGap.WebUI/Controllers/Report/ReportStartingController.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+             long organisationId = ControllerHelper.DecryptOrganisationIdOrThrow404(encryptedOrganisationId);
+             
+             ControllerHelper.ThrowIfUserAccountRetiredOrEmailNotVerified(User, dataRepository);
+             ControllerHelper.ThrowIfUserDoesNotHavePermissionsForGivenOrganisation(User, dataRepository, organisationId);
+             ControllerHelper.ThrowIfReportingYearIsOutsideOfRange(reportingYear);
+
+             var viewModel = new ReportStartingViewModel();
+
+             DraftReturn draftReturn = draftReturnService.GetDraftReturn(organisationId, reportingYear);
```
scl-softwire, line 44: I don't think this is needed - its only use is to set a particular property on the view model which isn't actually used by the view.

GenderPayGap.WebUI/Controllers/Report/ReportStartingController.cs
```diff
@@ -0,0 +1,66 @@
...
+             PopulateViewModel(viewModel, organisationId, reportingYear, draftReturn != null);
+
+             //Return submittedReturn = viewModel.Organisation.GetReturn(reportingYear);
```
scl-softwire, line 47: Commented-out code can be deleted

GenderPayGap.WebUI/Controllers/Report/ReportStartingController.cs
```diff
@@ -0,0 +1,66 @@
...
+            DraftReturnService draftReturnService)
+        {
+            this.dataRepository = dataRepository;
+            this.draftReturnService = draftReturnService;
+        }
+        
+        #region GET
+        
+        [HttpGet("{encryptedOrganisationId}/reporting-year-{reportingYear}/report/start")]
+        public IActionResult ReportFiguresGet(string encryptedOrganisationId, int reportingYear)
```
scl-softwire, line 34: This method name doesn't sound related to what it's actually doing

GenderPayGap.WebUI/Models/ReportStarting/ReportStartingViewModel.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+
+namespace GenderPayGap.WebUI.Models.ReportStarting
+{
+    public class ReportStartingViewModel
+    {
+        [BindNever /* Output Only - only used for sending data from the Controller to the View */]
```
scl-softwire, line 8: The [BindNever] attributes aren't needed here since this view model isn't used for a form submission

GenderPayGap.WebUI/Models/ReportStarting/ReportStartingViewModel.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace GenderPayGap.WebUI.Models.ReportStarting
+{
+    public class ReportStartingViewModel
+    {
+        [BindNever /* Output Only - only used for sending data from the Controller to the View */]
+        public Database.Organisation Organisation { get; set; }
+        [BindNever /* Output Only - only used for sending data from the Controller to the View */]
+        public int ReportingYear { get; set; }
+        
+        public bool IsEditingSubmittedReturn { get; set; }
```
scl-softwire, line 13: IsEditingSubmittedReturn and IsEditingForTheFirstTime don't appear to be used and can be removed from the model

GenderPayGap.WebUI/Views/ReportStarting/ReportStarting.cshtml
```diff
@@ -0,0 +1,84 @@
+﻿@using GovUkDesignSystem.GovUkDesignSystemComponents
+@using GovUkDesignSystem
+@model GenderPayGap.WebUI.Models.ReportStarting.ReportStartingViewModel
+
+@{
+    ViewBag.Title = $"Report gender pay gap data - reporting year  - Gender pay gap service"; // TODO: insert snapshot date
```
scl-softwire, line 6: I think this title is missing the actual year - other report pages include this information (is this what the TODO is talking about?)

GenderPayGap.WebUI/Views/ReportStarting/ReportStarting.cshtml
```diff
@@ -0,0 +1,84 @@
...
+        {
+            Html = @<text>
+                       Start now
+                       <svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" aria-hidden="true" focusable="false">
+                           <path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z"/>
+                       </svg>
+                    </text>,
+            Classes = "govuk-button--start govuk-!-margin-top-6",
+            Href = @Url.Action("ReportSizeOfOrganisationGet", "ReportSizeOfOrganisation", actionValues)
+        });
```
scl-softwire, line 77: Rogue semicolon here!

GenderPayGap.WebUI/Views/ManageOrganisations/OrganisationReportsTable.cshtml
```diff
@@ -112,7 +112,7 @@
                 {
                     Html = @<text>
                                 <a loadtest-id="create-report-@viewModel.ReportingYear"
-                                    href="@Url.Action("ReportOverview", "ReportOverview", new { encryptedOrganisationId, reportingYear = viewModel.ReportingYear, canTriggerLateSubmissionWarning = true})">
+                                    href="@Url.Action("ReportFiguresGet", "ReportStarting", new { encryptedOrganisationId, reportingYear = viewModel.ReportingYear})">
```
scl-softwire, line 115: Is is true that this should always link to the starting page? I'm not au fait with the new journey, but I would have expected that if a report has already been started, then this should either go to the point in the journey that they need to continue from, or to the overview page. As it is, if someone's already started a report and they want to come back to it later, this takes them right back to the start!

Airk0n: I discussed this and we think it should go to the "review page" I'm thinking I check in the controller if a draft exists and send them to the review page if it does. Does that sound like a reasonable approach or would you suggest doing that condition in the view?

scl-softwire: That approach sounds good to me!

Airk0n:
> Some local changes in inline comments. More generally:
> 
> 1. I find it weird that there are is one controller for each of the Report pages, and that these controllers are all broadly similar, rather than there being a single controller which is responsible for all of the pages. But it looks like it's been that way for a while and this ticket isn't the time to change it.
> 2. Consistency of Back buttons on report pages. At the moment if you start a report, hit "Start Now" to get to the headcount question, and then press "Back", you get taken to the report overview page, which is not actually the page you came from. Given that there are changes to be made to the other report pages, perhaps not fixable now, but certainly something to keep in mind.

Yeah we had a bit of a discussion about where to place controllers, I'm not overly happy with how they're separated, it might be worth adding a ticket to separate related user journeys into controllers or something to that effect.

I believe there's changes to be made to the headcount question page which will give us an opportunity to sort that, my assumption is that the back button will take you through the reporting journey in reverse.

GenderPayGap.WebUI/Controllers/Report/ReportStartingController.cs
```diff
@@ -0,0 +1,60 @@
...
+        
+        private readonly IDataRepository dataRepository;
+        private readonly DraftReturnService draftReturnService;
+
+        public ReportStartingController(
+            IDataRepository dataRepository,
+            DraftReturnService draftReturnService)
+        {
+            this.dataRepository = dataRepository;
+            this.draftReturnService = draftReturnService;
```
scl-softwire, line 28: Now that the call to this service has gone, I think it can be removed entirely from the class.

scl-softwire (changes requested):

PR #502, opened 2022-11-23
GPG-920: styling and content changed to match new designs. Links changed

Airk0n:
This ticket is concerned with changing styling and content of the reporting gender pay difference page.

**The main noteworthy changes:**
- "Step 2 of 5" above Heading.
- The difference is moved to the top
- Inset text used.
- Back used instead of breadcrumbs
- The difference between hourly pay and bonus pay is more clearly defined.
- Generally less text.

The design 
image

The implemented version
image

The old design
image

### DIFF
--- a/GenderPayGap.WebUI/Views/ReportFigures/ReportFigures.cshtml
+++ b/GenderPayGap.WebUI/Views/ReportFigures/ReportFigures.cshtml
@@ -17,35 +17,11 @@
 }
 
 @section BeforeMain {
-    @{
-        var crumbs = new List<CrumbViewModel>
-        {
-            new CrumbViewModel
-            {
-                Text = "Manage Employers",
-                Href = Url.Action("ManageOrganisationsGet", "ManageOrganisations")
-            },
-            new CrumbViewModel
-            {
-                Text = Model.Organisation.OrganisationName,
-                Href = Url.Action("ManageOrganisationGet", "ManageOrganisations", new {encryptedOrganisationId})
-            },
-            new CrumbViewModel
-            {
-                Text = $"{formattedReportingYears} reporting year",
-                Href = Url.Action("ReportOverview", "ReportOverview", new { encryptedOrganisationId, reportingYear = Model.ReportingYear})
-            },
-            new CrumbViewModel
-            {
-                Text = "Report figures"
-            }
-        };
-    }
-
-    @(await Html.GovUkBreadcrumbs(new BreadcrumbsViewModel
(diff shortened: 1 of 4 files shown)


GenderPayGap.WebUI/Views/ReportFigures/ReportFigures.cshtml
```diff
@@ -59,28 +35,26 @@
 
             <fieldset class="govuk-fieldset">
                 <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
-                    <h1 class="govuk-fieldset__heading">
-                        Enter your gender pay gap data for snapshot date @(Model.SnapshotDate.ToString("dd MMMM yyyy"))
+                    <h2 class="govuk-caption-l">Step 2 of 5</h2>
```
Airk0n, line 38: Looking at the W3C standards for accessibility it was ok with with a H2 after a H1 as I'm essentially starting a new section, an argument could be made for the Step 2 of 5 being a H1 instead and the main large text being a H2 instead.

davidm-m: I feel like the "Step 2 of 5" isn't a heading at all, just informative text. Probably the same with "for <organisation>". Can we make them spans instead of headings? And maybe the organisation name should be a `govuk-caption-m` instead of `heading`

GenderPayGap.WebUI/Views/ReportFigures/ReportFigures.cshtml
```diff
@@ -59,28 +35,26 @@
...
+                    <h1 class="govuk-heading-xl">
+                        Enter your gender pay gap data
```
davidm-m, line 40: The designs have this as "Enter gender pay gap data" (i.e. without the "your")

GenderPayGap.WebUI/Views/ReportFigures/ReportFigures.cshtml
```diff
@@ -107,11 +81,50 @@
                     Func<object, object> percentageText = @<text>
                                                               <span class="govuk-label govuk-!-margin-left-1" style="display: inline">%</span>
```
davidm-m, line 82: Can we use `govuk-!-display-inline` instead of setting a style on the element?

davidm-m (changes requested):
Looking very good, just a couple of things.

davidm-m (approved):

